PR #79, opened 2019-09-30
CMake: Add long long for size_t and don't use update-version.sh

1480c1:
Mingw-w64 has size_t as `unsigned long long` for win64

Remove reliance sh, use cmake to generate version.h


libvisual/CMakeLists.txt
```diff
@@ -102,6 +103,13 @@ ELSEIF(SIZEOF_SIZE_T EQUAL SIZEOF_INT)
 ELSEIF(SIZEOF_SIZE_T EQUAL SIZEOF_LONG)
   SET(VISUAL_SIZE_TYPE long)
   SET(VISUAL_SIZE_T_FORMAT "lu")
+ELSEIF(SIZEOF_SIZE_T EQUAL SIZEOF_LONG_LONG)
+  SET(VISUAL_SIZE_TYPE "long long")
+  IF(WIN32)
```
hartwork: I think this needs to be `MSVC` rather than `WIN32`?

libvisual/CMakeLists.txt
```diff
@@ -102,6 +103,13 @@ ELSEIF(SIZEOF_SIZE_T EQUAL SIZEOF_INT)
...
+    SET(VISUAL_SIZE_T_FORMAT "I64u")
+  else()
```
hartwork: Would be nice to have `else` uppercase for consistency

libvisual/CMakeLists.txt
```diff
@@ -179,14 +187,29 @@ ENDIF()
...
+
+IF(EXISTS ${BINARY_DIR}/version.h)
+  FILE(STRINGS ${BINARY_DIR}/version.h OLD_REVISION REGEX "#define LV_REVISION")
+  STRING(REGEX MATCH \".*\" OLD_REVISION ${OLD_REVISION})
+  STRING(REGEX MATCHALL "[^\"][0-9A-z]+" OLD_REVISION ${OLD_REVISION})
+ELSE()
+  SET(OLD_REVISION "")
+ENDIF()
+
+IF(NOT LV_REVISION EQUAL OLD_REVISION)
```
hartwork: This seems to be an optimisation to not re-generate a file with identical content every time someone runs CMake — true? If that's the case I wonder if that's worth the increase in complexity. If we keep it, let's add a comment why we need it.

libvisual/CMakeLists.txt
1480c1: No since this type also applies to mingw-w64 gcc

libvisual/CMakeLists.txt
1480c1: I'm not sure since that's what the original version.sh was doing so I was just replicating it. I'm not sure if it's necessary because I haven't seen someone use the same build folder contents between builds yet

libvisual/CMakeLists.txt
hartwork: Seems like you're right:
```console
# cat main.c 
#include <stdio.h>

int main() {
    const unsigned long long llu = 0x1234567812345678lu;
    printf("sizeof(long long): %u\n", sizeof(long long));
    printf("         expected: 1311768465173141112\n");
    printf("             I64u: %I64u\n", llu);
    return 0;
}

# i686-w64-mingw32-gcc -Wall -Wextra main.c && wine ./a.exe 
sizeof(long long): 8
         expected: 1311768465173141112
             I64u: 1311768465173141112
```

libvisual/CMakeLists.txt
```diff
@@ -185,14 +193,29 @@ ENDIF()
...
-  "${PROJECT_SOURCE_DIR}/update-version.sh.in"
-  "${PROJECT_BINARY_DIR}/update-version.sh"
-  @ONLY
-)
-ADD_CUSTOM_TARGET(update-version ALL
-  COMMAND /bin/sh ${PROJECT_BINARY_DIR}/update-version.sh
-)
+
+EXECUTE_PROCESS(COMMAND ${GIT_EXECUTABLE} describe --always --dirty
+  OUTPUT_VARIABLE LV_REVISION OUTPUT_STRIP_TRAILING_WHITESPACE)
```
hartwork: Because this command is no longer in `ADD_CUSTOM_TARGET` but run unconditionally, this will get users building from tarball stderr output like this (I just tried):
```
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```
It would be nice to not do that, to maybe not call Git in that case, at all. Maybe we can call Git if the version file is missing?

libvisual/CMakeLists.txt
```diff
@@ -185,14 +193,29 @@ ENDIF()
...
+IF(NOT LV_REVISION)
+  SET(LV_REVISION "unknown")
+ENDIF()
+
+IF(EXISTS ${BINARY_DIR}/version.h)
+  FILE(STRINGS ${BINARY_DIR}/version.h OLD_REVISION REGEX "#define LV_REVISION")
```
hartwork: It reads `PROJECT_BINARY_DIR` (not `BINARY_DIR`) further down and from what I tried, `PROJECT_BINARY_DIR` worked for me but with `BINARY_DIR` the file is not found at least for me.

libvisual/CMakeLists.txt
```diff
@@ -185,14 +193,29 @@ ENDIF()
...
+  STRING(REGEX MATCH \".*\" OLD_REVISION ${OLD_REVISION})
+  STRING(REGEX MATCHALL "[^\"][0-9A-z]+" OLD_REVISION ${OLD_REVISION})
```
hartwork: Capital `A` and lowercase `z` in `A-z` looks unintended and did not work in my tests.
Also, the use of `MATCHALL` seems to be a problem with Git output like `fa8021a6-dirty`. What works for me:
```
STRING(REGEX MATCH "[^\"][0-9a-z]+" OLD_REVISION ${OLD_REVISION})
                 ^^            ^
```
I would suggest we apply the same thing to `LV_REVISION` as well so that the `-dirty` in `fa8021a6-dirty` does not trigger re-generation every time.

libvisual/CMakeLists.txt
hartwork: Okay

libvisual/CMakeLists.txt
1480c1: Forgot to add `IF(GIT_FOUND)` to limit that. Should the default stay as unknown?

libvisual/CMakeLists.txt
1480c1: I literally don't know how I forgot `PROJECT_`. I even used it right below

libvisual/CMakeLists.txt
1480c1: Should we not use `--dirty` to append `-dirty` to the revision?

hartwork: I'm not sure. We could also make our regex allow `-` and then the `dirty` would make it's way through on both sides.

libvisual/CMakeLists.txt
```diff
@@ -185,14 +193,30 @@ ENDIF()
...
 FIND_PACKAGE(Git)
-CONFIGURE_FILE(
-  "${PROJECT_SOURCE_DIR}/update-version.sh.in"
-  "${PROJECT_BINARY_DIR}/update-version.sh"
-  @ONLY
-)
-ADD_CUSTOM_TARGET(update-version ALL
-  COMMAND /bin/sh ${PROJECT_BINARY_DIR}/update-version.sh
-)
+IF(GIT_FOUND)
```
hartwork: I'm not sure if this check works as intended. This would evaluate true for Git available but no repository below our feet. Git would be run an produce error output then. Is that what we want?

hartwork:
PS: shall we drop file `update-version.sh.in` from the repository as well?

1480c1:
I'm not sure yet, I was keeping it in tree for now until I confirm that this works completely

libvisual/CMakeLists.txt
1480c1: I've added `ERROR_QUIET` so that whatever error git shows, it won't be passed to the user and I think this is what we want since if we don't know what revision we are on, we would want to set it to the default

hartwork: The latest version up here does not seem to contain `ERROR_QUIET`. Did you push the latest version?

hartwork:
PS: Please rebase against latest master so that CI has a chance to go green. There was a regression on `master`.

libvisual/CMakeLists.txt
1480c1: I seemed to have forgot to press enter before I left

PR #85, opened 2019-12-22
CMake: Change minimum required to 2.8.12

1480c1:
Drops support for ancient versions. Even Centos 7 has 2.8.12 in their package manager.

Makes it easier to work with CMake without needing to worry about what's missing in such an old version


1480c1:
Done

PR #96, opened 2020-01-13
lv_cpu: Use Rip if on 64-bit Windows

1480c1:
Fixes #77

PR #97, opened 2020-02-03
actions: Add mingw-w64 build

1480c1:
As mentioned in https://github.com/Libvisual/libvisual/pull/96

Planning to try to add a msvc build

Currently will failed due to #96 and because `std::printf` is not a valid identifier for mingw-w64 since it's defined to `__printf__`

https://github.com/1480c1/libvisual/runs/424177417?check_suite_focus=true


hartwork:
What do you suggest we do about the printf problem? Maybe it would be good to fix that with a dedicated commit as part of this pull request?

1480c1:
Added basic stuff for libvisual dir

With #96 applied https://github.com/1480c1/libvisual/runs/425945846?check_suite_focus=true

1480c1:
Should I merge #96 with this pr?

hartwork:
> Should I merge #96 with this pr?

I was wondering the same already. Yes! Rebase on #96 please and then #97 has an actual chance at going green and #96 will be auto-marked as merged by GitHub once we merge #97. Sounds great to me — what do you think?

1480c1:
Now testing building the plugin

1480c1:
Perhaps I should wait before trying to compile the plugin since it seems there's some permissive errors due to storing a pointer into a long

.github/workflows/windows.yml
```diff
@@ -0,0 +1,93 @@
...
+          echo "::add-path::${env:GITHUB_WORKSPACE}/msys64/usr/bin"
+
+      - shell: cmd
+        if: steps.cache-msys2.outputs.cache-hit != 'true'
+        run: bash -lc 'exit'
+
+      - name: Update Pacman
+        run: pacman -Syy --needed --ask=20 --noconfirm
+
+      - name: Cache pacman
```
hartwork: This is about downloads for individual packages rather than the package database, right? My vote for making the name show.

.github/workflows/windows.yml
```diff
@@ -0,0 +1,93 @@
...
+          cmake -B build_lv -S libvisual -G"Unix Makefiles" -DCMAKE_BUILD_TYPE=Debug -DBUILD_SHARED_LIBS=OFF -DENABLE_DOCS=yes -DENABLE_NLS=OFF -DENABLE_TESTS=yes -DCMAKE_INSTALL_PREFIX="${env:GITHUB_WORKSPACE}/msys64/mingw64"
+
+      - name: '[LV] Run "make"'
+        run: |
+          make -C build_lv -j2 VERBOSE=1
+
+      - name: '[LV] Run "make install"'
+        run: |
+          make -C build_lv install
+          find ${env:GITHUB_WORKSPACE}/msys64/mingw64/bin/lv-tool ${env:GITHUB_WORKSPACE}/msys64/mingw64/include/libvisual-0.5 ${env:GITHUB_WORKSPACE}/msys64/mingw64/lib/libvisual-0.5* ${env:GITHUB_WORKSPACE}/msys64/mingw64/lib/pkgconfig/libvisual-0.5.pc | sort
```
hartwork: This whitelist seems to go against the idea of getting a listing of all files installed, including files that we are not expecting to be installed. The existing GitHub action uses a custom isolated `CMAKE_INSTALL_PREFIX`. That way we would not mix with msys64 and would not need a whitelist like that. Does it cause problems that I am not aware of?

libvisual/tools/lv-tool/lv-tool.cpp
```diff
@@ -98,7 +98,7 @@ namespace {
   /** print info about libvisual plugin */
   void print_plugin_info(VisPluginInfo const& info)
   {
-      std::printf("Plugin: \"%s\" (%s)\n"
+      printf("Plugin: \"%s\" (%s)\n"
```
hartwork: Does the file need addition of `using namespace std;`?

PS: It would be cool if the related commit's message could explain why that change is necessary.

libvisual/tools/lv-tool/display/stdout_sdl_driver.cpp
```diff
@@ -30,6 +30,9 @@
 
 #include <SDL/SDL.h>
 #include <GL/gl.h>
+#ifndef GL_BGR
+#include <GL/glext.h>
+#endif
```
hartwork: Is an unconditional include troublesome? If not, my vote for dropping `#ifndef GL_BGR`.

PS: Would be great if the related commit could mention how `GL_BGR` is defined only in some cases but not in others.

libvisual/libvisual/lv_cpu.c
```diff
@@ -133,7 +133,11 @@ static void sigfpe_handler_sse (int signal, struct sigcontext sc)
 LONG CALLBACK win32_sig_handler_sse (EXCEPTION_POINTERS* ep)
 {
 	if (ep->ExceptionRecord->ExceptionCode == EXCEPTION_ILLEGAL_INSTRUCTION) {
+#ifdef _AMD64_
```
hartwork: I have grepped the code base for `_AMD64_` now and did not find any other use. To blend in, we should probably give `VISUAL_OS_WIN32` a sibling `VISUAL_OS_WIN64` and check for that, instead? The fact that I don't find `_AMD64_` on https://sourceforge.net/p/predef/wiki/Architectures/ and that I ran into https://stackoverflow.com/a/17761728/11626624 makes me also wonder if `_AMD64_` the best choice.

libvisual/CMakeLists.txt
```diff
@@ -146,8 +146,8 @@ ENDIF()
 
 # Build profiles
 # Standard
-SET(CMAKE_C_FLAGS "${OPT_C_FLAGS} ${PROFILE_C_FLAGS} -std=c99 -Wall -Wextra -Wunused ${WERROR_C_FLAGS} -fstrict-aliasing")
-SET(CMAKE_CXX_FLAGS "${OPT_C_FLAGS} ${PROFILE_C_FLAGS} ${CXX11_FLAGS} -Wall -Wextra -Wunused ${WERROR_C_FLAGS} -fstrict-aliasing")
+SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OPT_C_FLAGS} ${PROFILE_C_FLAGS} -std=c99 -Wall -Wextra -Wunused ${WERROR_C_FLAGS} -fstrict-aliasing")
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OPT_C_FLAGS} ${PROFILE_C_FLAGS} ${CXX11_FLAGS} -Wall -Wextra -Wunused ${WERROR_C_FLAGS} -fstrict-aliasing")
```
hartwork: Is this about supporting flags provided by the user? If so, I'd expect them to go last so they can override `${OPT_C_FLAGS}` and friends. Why this way around? What was this addition needed for?

.github/workflows/windows.yml
```diff
@@ -0,0 +1,93 @@
...
+          rm msys2.tar
+
+      - name: Add Msys2 to Path
+        run: |
+          echo "::add-path::${env:GITHUB_WORKSPACE}/msys64/mingw64/bin"
+          echo "::add-path::${env:GITHUB_WORKSPACE}/msys64/usr/bin"
+
+      - shell: cmd
+        if: steps.cache-msys2.outputs.cache-hit != 'true'
+        run: bash -lc 'exit'
```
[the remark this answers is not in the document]
1480c1: The main reason is to setup msys2 as the first time it runs, it needs to setup itself and initialize the home directory and pacman

You can see an output of this here https://github.com/1480c1/libvisual/runs/424361604?check_suite_focus=true

Should I name it to something like `Initialize msys2`?

.github/workflows/windows.yml
```diff
@@ -0,0 +1,93 @@
...
+
+      - name: Update Pacman
+        run: pacman -Syy --needed --ask=20 --noconfirm
```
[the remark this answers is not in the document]
1480c1: I just left `--ask=20` in there since I split this command off the other pacman command to update the packages themselves to see if that would influence the caching job right below it. I'm not aware of any instances where `-Syy` would fail or ask for confirmation, so I think I can just remove it

.github/workflows/windows.yml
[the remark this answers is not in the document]
hartwork: I see! `Initialize msys2` sounds pretty good to me.

.github/workflows/windows.yml
```diff
@@ -0,0 +1,93 @@
...
+        run: pacman -Suu --needed --ask=20 --noconfirm
+
+      - name: Install packages
+        env:
+          pkg: orc pkg-config cmake gcc libpng SDL gettext
+        run: pacman -S --needed --ask=20 --noconfirm make $env:pkg.Split(' ').ForEach({"mingw-w64-x86_64-$_"})
+
+      - name: "[LV] Run CMake"
+        run: |
+          cmake -B build_lv -S libvisual -G"Unix Makefiles" -DCMAKE_BUILD_TYPE=Debug -DBUILD_SHARED_LIBS=OFF -DENABLE_DOCS=yes -DENABLE_NLS=OFF -DENABLE_TESTS=yes -DCMAKE_INSTALL_PREFIX="${env:GITHUB_WORKSPACE}/msys64/mingw64"
```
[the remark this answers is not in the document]
1480c1: Yep, `MinGW Makefiles` generates makefiles for mingw32-make and doesn't allow `sh.exe` to be in path. So it's either `MSYS Makefiles`, `Unix Makefiles` ,or `Ninja`, but I opted for `Unix Makefiles` to match the other other one and for color since `Ninja` seems to not have color a lot of the times. `MSYS Makefiles` uses the `/c/...` format which I don't prefer since it makes it a bit hard to read

```diff
-CMAKE_COMMAND = /D/msys64/mingw64/bin/cmake.exe #MSYS
+CMAKE_COMMAND = D:/msys64/mingw64/bin/cmake.exe #Unix
```

Both Unix and MSYS run fine with make.exe from any shell (at least msys2 and powershell for me)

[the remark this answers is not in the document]
hartwork: Thanks for the clarification!

.github/workflows/windows.yml
```diff
@@ -0,0 +1,93 @@
...
+        id: cache-msys2
+        uses: actions/cache@v1.1.0
+        with:
+          path: msys64
+          key: ${{ runner.os }}-msys64
+
+      - name: Get Msys2
+        if: steps.cache-msys2.outputs.cache-hit != 'true'
+        run: |
+          Invoke-WebRequest -Outfile msys2.tar.xz -Uri "https://github.com/msys2/msys2-installer/releases/download/nightly-x86_64/msys2-base-x86_64-latest.tar.xz"
```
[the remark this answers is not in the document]
1480c1: The main reason why is because msys2 works on a rolling build so we could either get one from http://repo.msys2.org/distrib/ and update all of the packages, or use a `nightly` build that  I assume is built the same way as the regular distributions to reduce the amount we would have to download and update.

Plus, since it's from github's servers, I doubt that we would have issues downloading the tar.xz on the account that if github is down, their actions would be down also.

I'm not sure if there's a good way to reliably `pin` the version with pacman.

Perhaps the best way to make sure these are reproducible would be to upload the msys64 folder and reuse that forever without updating pacman (similar to what the cache job is doing right now).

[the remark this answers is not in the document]
hartwork: Good points, let's keep things this way then.

.github/workflows/windows.yml
```diff
@@ -0,0 +1,93 @@
...
+        run: pacman -Syy --needed --ask=20 --noconfirm
+
+      - name: Cache pacman
+        id: cache-pacman
+        uses: actions/cache@v1.1.0
+        with:
+          path: msys64/var/cache/pacman/pkg
+          key: ${{ runner.os }}-pacman-${{ hashFiles('msys64/var/lib/pacman/sync/mingw64.db') }}
+          restore-keys: |
+            ${{ runner.os }}-pacman-
```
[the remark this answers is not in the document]
1480c1: I'm mainly using this in hope that whenever a package is updated and downloaded, github will recache the packages instead of keeping stale ones like travis and appveyor, so this is my attempt at it, I'm still not sure on how `action/cache` works with their keys.

I can remove it since it always seems to hit

.github/workflows/windows.yml
```diff
@@ -0,0 +1,93 @@
...
+
+      - name: Update Msys2 Packages
+        run: pacman -Suu --needed --ask=20 --noconfirm
```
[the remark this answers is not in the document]
1480c1: Probably it's me that is missing something, currently cloning pacman source so I figure out more about it.

But yeah, I added double u to just in case there's a downgrade because of bugs or incompatibilities since that has happened before

libvisual/tools/lv-tool/lv-tool.cpp
1480c1: I don't think it should since it should pull the c linkage version of `printf` 

I can amend the commit message

libvisual/tools/lv-tool/display/stdout_sdl_driver.cpp
1480c1: I mainly added an `#ifndef` because I only readed it for `GL_BGR` since it seems that for msys2, `GL_BGR` is not defined in `GL/gl.h` as the rest of GL stuff and didn't want to include it if necessary.

I can remove the conditionals and amend the message

libvisual/libvisual/lv_cpu.c
1480c1: I mainly found `_AMD64_` within windows' headers as the guards for if `RIP` is present in the struct, so I'm assuming it's mainly an internal define. I could try to add `VISUAL_OS_WIN64` since it might also be helpful to distinguish 64-bit too

libvisual/CMakeLists.txt
1480c1: This addition was originally there since I needed it for adding another lib, but it was eventually nullified by a different CMake modification. I just added it to the beginning due to a habit with other CMakeLists, I will move it over

libvisual/tools/lv-tool/lv-tool.cpp
hartwork: > I don't think it should since it should pull the c linkage version of `printf`

```console
# echo $'#include <cstdio>\n\nint main() { printf("Hello\\n"); return 0; }' > main.cxx
# gcc -Wall -Wextra -pedantic -std=c++98 main.cxx
```
Seems like you're right — I was expecting some warning or so at least. Okay :smiley: 

> I can amend the commit message

As you like

1480c1: I will rebase amend them after I finalize any more changes

libvisual/libvisual/lv_cpu.c
hartwork: PS: If 64bit detection in CMake to build `VISUAL_OS_WIN64` turns out hard, I would vote for `_WIN64` over `_AMD64_`: it seems more robust / official to me (and the surrounding code is Windows-only already).

1480c1: It seems easy enough to do
```cmake
...
ELSEIF(WIN32)
  SET(VISUAL_OS_WIN32 yes)
  SET(VISUAL_WITH_MINGW ${MINGW})
  SET(VISUAL_WITH_CYGWIN ${CYGWIN})
  IF(VISUAL_WITH_CYGWIN)
    SET(VISUAL_OS_POSIX yes)
  ENDIF()
  IF(${CMAKE_SIZEOF_VOID_P} EQUAL 8)
    SET(VISUAL_OS_WIN64 yes)
  ENDIF()
ELSE()
...
```

libvisual/tools/lv-tool/display/stdout_sdl_driver.cpp
hartwork: Maybe?:
```c++
#include <GL/glext.h>  // e.g. for GL_BGR on MinGW
```

1480c1: I used ` // For GL_BGR in systems where it's not defined in GL/gl.h` since I didn't see your message in time

libvisual/libvisual/lv_cpu.c
hartwork: Good idea

libvisual/tools/lv-tool/display/stdout_sdl_driver.cpp
hartwork: Good enough for sure, thanks!

.github/workflows/windows.yml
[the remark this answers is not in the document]
1480c1: In lib/libalpm/sync.c

```c
		if(enable_downgrade) {
			/* check IgnorePkg/IgnoreGroup */
			if(alpm_pkg_should_ignore(handle, spkg)
					|| alpm_pkg_should_ignore(handle, lpkg)) {
				_alpm_log(handle, ALPM_LOG_WARNING, _("%s: ignoring package downgrade (%s => %s)\n"),
						lpkg->name, lpkg->version, spkg->version);
			} else {
				_alpm_log(handle, ALPM_LOG_WARNING, _("%s: downgrading from version %s to version %s\n"),
						lpkg->name, lpkg->version, spkg->version);
				return 1;
			}
		} else {
			alpm_db_t *sdb = alpm_pkg_get_db(spkg);
			_alpm_log(handle, ALPM_LOG_WARNING, _("%s: local (%s) is newer than %s (%s)\n"),
					lpkg->name, lpkg->version, sdb->treename, spkg->version);
		}
```

I think they ignore the ask option since they don't have anything with `config->ask` (where they store the `(unsigned int)20` into) nearby or in any of the functions

(It's so hard to read someone else's code from scratch.)

PR #107, opened 2022-11-21
[0.4.x] libvisual: Allow strict aliasing (-Werror=strict-aliasing)

hartwork:
First raised at https://bugs.gentoo.org/859922

Note that `master` already has that problem fixed.


thesamesam (approved):
Looks fine, thanks.

hartwork:
@thesamesam thanks for the review! :+1:

Issue #33, opened 2013-01-05
LV: Visualization lags when using PulseAudio


kaixiong:
The (incorrect) threading code in input_pulseaudio was recently removed in attempt to fix the weird issue where rendered frames appear to flip back and forth.

This reduced the implementation to a single blocking read with a single pa_simple_read() of 1024 samples (at 44,100Hz) followed by a frame render per main loop iteration and worked well on my test machine using PulseAudio 1.1 (Ubuntu 12.04 x64) but stutters horribly on PulseAudio 2.1 (tested on Ubuntu 12.10 x64). 

Although I haven't measured, I believed pa_simple_read() is taking longer to return than it needs to be, and is incredibly inconsistent timing wise despite reading the same amount of samples every frame. More investigation is needed. 

We will likely have to either restore threading code or use the complex API. Both approaches kind of suck.

EDIT: apparently this isn't limited to pulseaudio input.

hartwork:
A quick note that I have reading from the microphone using the [Asynchronous API of PulseAudio](https://freedesktop.org/software/pulseaudio/doxygen/async.html) working at https://github.com/hartwork/audio_pump_demo . I could port that approach to libvisual `input_pulseaudio.c` as needed.

hartwork:
There is pull request #149 now. Not that it's for `0.4.x` rather than `master`, `master` is planned to be next.

PR #149, opened 2023-01-04
[0.4.x] libvisual-plugins: Add PulseAudio input plugin (based on PulseAudio's Asynchronous API)

hartwork:
This a backport from `master` to `0.4.x` (that didn't have any PulsesAudio input) followed by a migration from PulseAudio's [Simple API](https://freedesktop.org/software/pulseaudio/doxygen/simple.html) to PulseAudio's [Asynchronous API](https://freedesktop.org/software/pulseaudio/doxygen/async.html) to make it work in practice. Forward-porting this migration to `master` will be a follow-up pull request.

Related to #33.

### DIFF
--- a/libvisual-plugins/configure.ac
+++ b/libvisual-plugins/configure.ac
@@ -95,6 +95,21 @@ build_input_plugins=""
 build_actor_plugins=""
 build_morph_plugins=""
 
+dnl PulseAudio
+AC_ARG_ENABLE([pulseaudio], AS_HELP_STRING([--disable-pulseaudio],
+	[Do not build PulseAudio input plugin @<:@default=enabled@:>@]),
+        [ENABLE_INPUT_PULSEAUDIO=$enableval],
+	[ENABLE_INPUT_PULSEAUDIO=yes])
+if test "$ENABLE_INPUT_PULSEAUDIO" = "yes"; then
+    PKG_CHECK_MODULES([LIBPULSE], [libpulse], [HAVE_LIBPULSE="yes"], [HAVE_LIBPULSE="no"])
+    if test "$HAVE_LIBPULSE" = "yes"; then
+        build_input_plugins="$build_input_plugins pulseaudio"
+    else
+        AC_MSG_WARN([*** Could not find libpulse of PulseAudio.
+                        The PulseAudio input plugin won't be built.])
+    fi
+fi
+
 dnl JACK
 AC_ARG_ENABLE([jack], AS_HELP_STRING([--disable-jack],
 	[Do not build jack input plugin @<:@default=enabled@:>@]),
@@ -572,6 +587,7 @@ AC_CONFIG_FILES([
 	plugins/input/jack/Makefile
 	plugins/input/debug/Makefile
 	plugins/input/portaudio/Makefile
+	plugins/input/pulseaudio/Makefile
 	plugins/morph/Makefile
 	plugins/morph/alphablend/Makefile
 	plugins/morph/tentacle/Makefile
@@ -589,6 +605,7 @@ LIBVISUAL-PLUGINS, PLUGINS FOR LIBVISUAL $VERSION
 
 Packages found:
 ALSA                     : $HAVE_ALSA
+PulseAudio               : $HAVE_LIBPULSE
 OpenGL                   : $HAVE_OPENGL
 GTK+                     : $HAVE_GTK
 
--- a/libvisual-plugins/plugins/input/Makefile.am
+++ b/libvisual-plugins/plugins/input/Makefile.am
@@ -1,4 +1,4 @@
 
 SUBDIRS = $(build_input_plugins)
 
-DIST_SUBDIRS = mplayer jack alsa debug portaudio
+DIST_SUBDIRS = mplayer jack alsa debug portaudio pulseaudio
--- a/libvisual-plugins/plugins/input/pulseaudio/Makefile.am
+++ b/libvisual-plugins/plugins/input/pulseaudio/Makefile.am
@@ -1,0 +1,8 @@
+## Process this file with automake to generate a Makefile.in
+
+input_plugin_LTLIBRARIES = input_pulseaudio.la
+
+input_pulseaudio_la_SOURCES = input_pulseaudio.c
+input_pulseaudio_la_CFLAGS = $(LIBVISUAL_CFLAGS) $(LIBPULSE_CFLAGS)
+input_pulseaudio_la_LIBADD = $(LIBVISUAL_LIBS) $(LIBPULSE_LIBS)
+input_pulseaudio_la_LDFLAGS = -module -avoid-version
--- a/libvisual-plugins/plugins/input/pulseaudio/input_pulseaudio.c
+++ b/libvisual-plugins/plugins/input/pulseaudio/input_pulseaudio.c
@@ -1,0 +1,179 @@
+/*    This file is part of Xmms2-libvisual.
+ *
+ *    Xmms2-libvisual is free software: you can redistribute it and/or modify
+ *    it under the terms of the GNU General Public License as published by
+ *    the Free Software Foundation, either version 3 of the License, or
+ *    (at your option) any later version.
+ *
+ *    Xmms2-libvisual is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with Xmms2-libvisual.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <libvisual/libvisual.h>
+#include "gettext.h"
+#include <stdlib.h>
+#include <string.h>
+#include <pulse/simple.h>
+#include <pulse/error.h>
+
+VISUAL_PLUGIN_API_VERSION_VALIDATOR
+
+#define SAMPLES 1024
+#define BUFFERS 2
+
+pa_sample_spec sample_spec = {
+    .format = PA_SAMPLE_S16LE,
+    .rate = 44100,
+    .channels = 2
+};
+
+typedef struct {
+    pa_simple *simple;
+    int16_t pcm_data[SAMPLES*2];
+} pulseaudio_priv_t;
+
+static int  inp_pulseaudio_init    (VisPluginData *plugin);
+static void inp_pulseaudio_cleanup (VisPluginData *plugin);
+static int  inp_pulseaudio_upload  (VisPluginData *plugin, VisAudio *audio);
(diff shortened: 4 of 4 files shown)


kaixiong (commented):
@hartwork This implementation is tricky. I am no expert on concurrency but there are two issues I have with this lock-free code.

1. Reads and writes to `uint64_t` are not guaranteed to be atomic across platforms - LV can run on 32-bit platforms. We should use C11 atomics here ([reference](https://en.cppreference.com/w/c/atomic)) or its equivalent and let the compiler deal with it.
2. In `upload()`, `priv->chunks_written` is read several times and then used for computation. The correctness is hard to ascertain since `on_input_stream_data()` can run any number of times in between and update its value. Perhaps the code could retrieve the value just once at the beginning.

I would like to see an informal proof of sorts that the code is correct.

hartwork:
> @hartwork This implementation is tricky. I am no expert on concurrency but there are two issues I have with this lock-free code.
> 
> 1. Reads and writes to `uint64_t` are not guaranteed to be atomic across platforms - LV can run on 32-bit platforms. We should use C11 atomics here ([reference](https://en.cppreference.com/w/c/atomic)) or its equivalent and let the compiler deal with it.

@kaixiong ouch. I would like to keep the 0.4.x branch C99 for now. I think the implications of what you are saying are that we could end up with a bigger value read from `priv->chunks_written` this way:
1. read half of `priv->chunks_written`,
2. increment `priv->chunks_written` in a way that changes both the upper and the lower half of bits, 
3. read other half of `priv->chunks_written`.

We could:
- a) consider a handmade workaround like https://groups.google.com/g/comp.programming.threads/c/TKa2ghGgNdA/m/aAQK4JTeAyAJ,
- b) consider dropping 32bit support, detect and reject 32 bit compiles,
- c) use [GCC's `__atomic_load`](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html), Clang seems to [have it too](https://llvm.org/docs/Atomics.html#libcalls-atomic), or
- d) switch to `uint32_t`, more on the math consequences below.

> 2. In `upload()`, `priv->chunks_written` is read several times and then used for computation. The correctness is hard to ascertain since `on_input_stream_data()` can run any number of times in between and update its value. Perhaps the code could retrieve the value just once at the beginning.

Good idea, done.

> I would like to see an informal proof of sorts that the code is correct.

Alright, let's see.

So imagine we have two athletes running on a single lane circular track, one is the reader, one is the writer.  What we want to achieve that at all times:
- The reader does not overtake the writer (to not get stale and duplicate reads).
- The writer does not overtake the reader (to not lose writes to nowhere and to not get corrupted reads).

Now let's first observe that if we would track the total (multi-lap) distance walked in bytes over time for per runner using a `uint64_t` counter, with <=176400 (`= 44100 * 2 * 2`) bytes coming in per second, that means that our application's process can run for 3 million years before `uint64_t` wraps around: `(2 ** 64 - 1) / 176400 / 60 / 60 / 24 / 365` is roughly 3,316,000. If instead we count chunks of 1024 bytes in `uint32_t`, we end up with a wrap after ~290 days of runtime (`(2 ** 32 - 1) * 1024 / 176400 / 60 / 60 / 24`).

Such a position counter solves the first problem: The reader just checks for the writer's walked distance before reading and advancing: no locks needed for that, done.

What can we do to ensure the writer does not run into the reader from behind?
Generally speaking, our options are:
- a) slow down the writer
- b) speed up the reader
- c) "stretch the ground" between writer and reader, add additional distance on ground level.

For slowing down the writer (a) the only way would be dropping samples.
For speeding up the writer (b) one way is skip-jumping (which means dropping samples) as a last resort and another is sprinting to where the writer is now by reading all waiting bytes at once.  This assumes that reading RAM is faster than 44100 frames per second.
While we don't stretch the ground dynamically (c) in the current implementation, the longer the track — i.e. the bigger the buffer — the more tolerance for temporary differences in speed — jittering — we get.  On a side note, an infinitely large buffer would solve that problem "for free".
I would like to note that unless `.render` blocks unusually long, `.upload` should call the reader at rendering FSP frequency, which is unlikely to ever fall below 1Hz, so that could be considered minimum speed guaranteed and maximum buffer size requirement in frames.

In practice, over here the reader fluctuates between being 0 to 4 chunks behind maximum while the buffer has space for ~350 chunks (`= 2 * 44100 / 256 + 1`), i.e. more than 100 times the reader lag.

PS: Things I have adjusted in the pull request today:
- Introduce `frozen_chunks_written` as you suggested above
- Mark `priv.chunks_written` as `volatile` because it's accessed by both threads
- Rename `SAMPLES` to `FRAMES`
- Add and improve in-code comments

What do you think?

hartwork:
@kaixiong I have made the code use `__atomic_*` now. I'm not 100% sure if `__ATOMIC_ACQUIRE` and `__ATOMIC_RELEASE` are the correct/best ordering values to use, my choice was mostly based on https://dmalcolm.fedorapeople.org/gcc/2015-08-31/rst-experiment/built-in-functions-for-memory-model-aware-atomic-operations.html#built-in-functions-for-memory-model-aware-atomic-operations . Also please note that some docs mention a first `size_t size` parameter that doesn't seem to exist in practice, neither with GCC 11 nor with Clang 16. Looking forward to what you think.

kaixiong:
@hartwork, so I did some research regarding atomicity and memory ordering and here's what I think:

- This code should be okay (barring `upload()` taking too long) as long as neither the compiler nor the multi-core CPU re-orders `chunks_written++` before the `visual_mem_copy()` call. If the compiler believes that that the function is completely independent of `chunks_written++`, then it's possible for 'visual_mem_copy()' to be ordered like this:

```c
if (priv->chunk_write_offset_bytes + round_nbytes < CHUNK_SIZE_BYTES) {
    ...
} else {
    ...
    priv->chunks_written++
}

// Possible re-order? Neither target, source nor round_nbytes changes in the if block!
// If so, upload() may start reading before the copy completes (or even before it begins)
visual_mem_copy(target, source, round_nbytes);
```

- The statement `chunks_written++ `, despite how it looks, involves 3 steps: read, modify and write. There may be a single hardware instruction that performs this atomically, but that is hardware specific. The most obvious example being a 32-bit CPU since it cannot operate on 64-bit values automatically. There may also be a 64-bit CPU that does not have an atomic 64-bit increment instruction.

- In GCC, volatile suppresses optimizations and prevents instruction re-ordering only with respect to other volatile accesses. It also does nothing to prevent CPU out-of-order executions. To prevent incorrect re-ordering by the compiler, a compiler-specific barrier has to be used. For GCC, it's `asm volatile ("" : : : "memory")`. To prevent incorrect CPU instruction re-ordering, a separate hardware fence instruction is needed.

The upshot is that reads and writes to `chunks_written`  must be explicitly made atomic (`volatile` is useless for our purpose), regardless of whether the code is executing on 32-bit or 64-bit machines.

Looking at the latest version of this PR:
* Acquire-Release semantics should be good enough. I had originally thought of using sequential consistent to keep things simple
* Always use `__atomic_load` and `__atomic_store` on `chunks_written` regardless of 32-bit or 64-bit. Leave it for the compiler to decide what to do i.e. insert locks/fences where appropriate, or refuse to compile. Do not build this plugin at all if the compiler does not support these two functions.

hartwork:
> * The statement `chunks_written++`, despite how it looks, involves 3 steps: read, modify and write. There may be a single hardware instruction that performs this atomically, but that is hardware specific. The most obvious example being a 32-bit CPU since it cannot operate on 64-bit values automatically. There may also be a 64-bit CPU that does not have an atomic 64-bit increment instruction.

@kaixiong my understanding is that we do not need atomic read-increment-write because we only have a single thread writing.

> * Always use `__atomic_load` and `__atomic_store` on `chunks_written` regardless of 32-bit or 64-bit. Leave it for the compiler to decide what to do i.e. insert locks/fences where appropriate, or refuse to compile. Do not build this plugin at all if the compiler does not support these two functions.

Okay, done, pushed.

kaixiong:
> @kaixiong my understanding is that we do not need atomic read-increment-write because we only have a single thread writing.

@hartwork, you're right.

> 
> > * Always use `__atomic_load` and `__atomic_store` on `chunks_written` regardless of 32-bit or 64-bit. Leave it for the compiler to decide what to do i.e. insert locks/fences where appropriate, or refuse to compile. Do not build this plugin at all if the compiler does not support these two functions.
> 
> Okay, done, pushed

Thanks :+1: . I'm checking if  `__atomic_thread_fence` is needed to prevent any re-ordering due to compiler and CPU.

I've been playing around with Godbolt and so far neither GCC nor Clang re-orders the `visual_mem_copy` on x86 and ARM. I've tried having `visual_mem_copy` as external blackbox, `memcpy`, and having it as a inlineable whitebox (implementing the functionality in a loop). Still need to see how/if CPU out-of-order executions can happen that causes `chunks_written` to be updated before the buffer copy is complete.

kaixiong:
@hartwork, so after more reading and checking Godbolt, I can confirm an explicit `__atomic_thread_fence()` is not needed. The Acquire and Release memory order applied to `chunks_written` accesses via calls to `__atomic_load` and `__atomic_store` will introduce the necessary load/store fences. The code should also work on 32-bit machines. As it is, the plugin implementation is good to go.

One possible but minor(?) improvement is the use of __atomic_fetch_add() to the `chunks_written` increment.

What I'd like to see though, is the check for __atomic_load() and __atomic_store() to be moved outside of the plugin code and into the build system as a compiler feature test. Whenever the feature test fails, then `input_pulseaudio` would be excluded from build like the other optional plugins. I'm okay with merging the existing code first, so I will approve the PR first.

kaixiong (approved):

hartwork:
> @hartwork, so after more reading and checking Godbolt, I can confirm an explicit `__atomic_thread_fence()` is not needed. The Acquire and Release memory order applied to `chunks_written` accesses via calls to `__atomic_load()` and `__atomic_store()` will introduce the necessary load/store fences. The code should also work on 32-bit machines. As it is, the plugin implementation is good to go.

@kaixiong cool!

> One possible but minor(?) improvement is the use of `__atomic_fetch_add()` for the `chunks_written` increment.

I think it's not needed because we only have a single thread writing.

> What I'd like to see though, is the check for `__atomic_load()` and `__atomic_store()` to be moved outside of the plugin code and into the build system as a compiler feature test. Whenever the feature test fails, then `input_pulseaudio` would be excluded from build like the other optional plugins. I'm okay with merging the existing code first, so I will approve the PR first.

I think there's two aspects to this:
- Catch all compile errors before compilation in the configure stage and with proper reporting; that's nice in general and will help people playing with more exotic compilers before running into actual compile errors.
- Our CI only supports GCC and Clang and it could be argued that that's all we support currently, anyway. If this is about `master` primarily, this could be tied to adding MSVC support to the CI later.

Merging now…

kaixiong:
@hartwork, leaving this link here in case you ever want to investigate the assembly code:
https://godbolt.org/z/GsT551d5s

PR #150, opened 2023-01-07
[0.4.x] libvisual: Backport lv-tool from master to 0.4.x

hartwork:


hartwork:
@kaixiong I have taught lv-tool to only upgrade the actor to `lv_gtltest` if it appears available in commit `lv-tool: Upgrade default actor plugin to "lv_gltest" if available` now.

kaixiong (commented):
@hartwork Code looks good but there seems to be some tab indentation mixed into lv_tool.cpp. I can't quite tell with Github's code viewer. The convention is four space characters for each indent level. Would be good to fix that.

We probably should try and get a usable editorconfig to make this easier to deal with. I run into the same problem now and then.

hartwork:
> @hartwork Code looks good but there seems to be some tab indentation mixed into lv_tool.cpp. I can't quite tell with Github's code viewer. The convention is four space characters for each indent level. Would be good to fix that.

@kaixiong you're right, it's all spaces on `master`…

```console
# grep $'^\t' ./libvisual/tools/lv-tool/lv-tool.cpp | wc -l
0
# grep '^ '   ./libvisual/tools/lv-tool/lv-tool.cpp | wc -l
511
```

…but some tab lines in here. I'll fix that :+1: 


> We probably should try and get a usable editorconfig to make this easier to deal with. I run into the same problem now and then.

It would be over-simplified to say that we have no tabs on `master`:

```console
# git grep -l $'^\t' | wc -l
290
# git grep -l $'^ ' | wc -l
681
```
For `0.4.x` it's half and half:
```console
# git grep -l $'^\t' | wc -l
373
# git grep -l $'^ ' | wc -l
361
```

I believe we need both an editorconfig and CI enforcing clang-format, e.g. similar to what [this `clang-format.yml`](https://github.com/hartwork/sdl_video_demo/blob/master/.github/workflows/clang-format.yml) is doing. The question to me is, when to best do mass-changes like that, e.g. because it affects/breaks cross-branch cherry-pickability and open pull requests. What do you think?

hartwork:
@kaixiong `./libvisual/tools/lv-tool/lv-tool.cpp` is now free of tabs.

kaixiong:
@hartwork You're right. That's mostly in the C source files and headers and have remained so due to the concern you just raised.

The worst violator in the C++ sections is lv_bin.cpp, lv_bin.h. The rest are kind of manageable.

kaixiong (approved):
@hartwork, all good!

hartwork:
> @hartwork, all good!

@kaixiong awesome, thanks for the review! :+1:

PR #152, opened 2023-01-07
[0.4.x] libvisual: Fix splitting of interleaved stereo

hartwork:
Previously, twice as many bytes as available were read and written (heap-buffer-overflow) and the code was not robust towards and odd number of samples.  Thanks to [AddressSanitizer](https://clang.llvm.org/docs/AddressSanitizer.html)!

PR #154, opened 2023-01-07
[0.4.x] libvisual-plugins: Fix ALSA input plugin

hartwork:
Should ideally be merged *after* pull request #152 or it will further increase the buffer overflow problem that #152 is fixing.

### DIFF
--- a/libvisual-plugins/plugins/input/alsa/input_alsa.c
+++ b/libvisual-plugins/plugins/input/alsa/input_alsa.c
@@ -51,7 +51,7 @@ static int inp_alsa_cleanup (VisPluginData *plugin);
 static int inp_alsa_upload (VisPluginData *plugin, VisAudio *audio);
 
 static const int  inp_alsa_var_btmul      = sizeof (short);
-static const char *inp_alsa_var_cdevice   = "hw:0,0";
+static const char *inp_alsa_var_cdevice   = "default";
 static const int  inp_alsa_var_samplerate = 44100;
 static const int  inp_alsa_var_channels   = 2;
 
--- a/libvisual-plugins/plugins/input/alsa/input_alsa.c
+++ b/libvisual-plugins/plugins/input/alsa/input_alsa.c
@@ -253,7 +253,7 @@ int inp_alsa_upload (VisPluginData *plugin, VisAudio *audio)
 		if (rcnt > 0) {
 			VisBuffer buffer;
 
-			visual_buffer_init (&buffer, data, rcnt, NULL);
+			visual_buffer_init (&buffer, data, rcnt * sizeof(data[0]) * inp_alsa_var_channels, NULL);
 
 			visual_audio_samplepool_input (audio->samplepool, &buffer, VISUAL_AUDIO_SAMPLE_RATE_44100,
 					VISUAL_AUDIO_SAMPLE_FORMAT_S16, VISUAL_AUDIO_SAMPLE_CHANNEL_STEREO);
--- a/libvisual-plugins/plugins/input/alsa/input_alsa.c
+++ b/libvisual-plugins/plugins/input/alsa/input_alsa.c
@@ -221,7 +221,8 @@ int inp_alsa_upload (VisPluginData *plugin, VisAudio *audio)
 {
 	int16_t data[PCM_BUF_SIZE];
 	alsaPrivate *priv = NULL;
-	int rcnt;
+	const snd_pcm_uframes_t frames_wanted = sizeof(data) / sizeof(data[0]) / inp_alsa_var_channels;
+	snd_pcm_sframes_t rcnt;
 	int i;
 
 	visual_log_return_val_if_fail(audio != NULL, -1);
@@ -247,7 +248,7 @@ int inp_alsa_upload (VisPluginData *plugin, VisAudio *audio)
 	}
 #endif
 	do {
-		rcnt = snd_pcm_readi(priv->chandle, data, PCM_BUF_SIZE / 2);
+		rcnt = snd_pcm_readi(priv->chandle, data, frames_wanted);
 
 		if (rcnt > 0) {
 			VisBuffer buffer;
--- a/libvisual-plugins/plugins/input/alsa/input_alsa.c
+++ b/libvisual-plugins/plugins/input/alsa/input_alsa.c
@@ -106,7 +106,7 @@ int inp_alsa_init (VisPluginData *plugin)
 
 	visual_object_set_private (VISUAL_OBJECT (plugin), priv);
 
-	if ((err = snd_pcm_open(&priv->chandle, strdup(inp_alsa_var_cdevice),
+	if ((err = snd_pcm_open(&priv->chandle, inp_alsa_var_cdevice,
 			SND_PCM_STREAM_CAPTURE, SND_PCM_NONBLOCK)) < 0) {
 		visual_log(VISUAL_LOG_CRITICAL, 
 			    _("Record open error: %s"), snd_strerror(err));
--- a/libvisual-plugins/plugins/input/alsa/input_alsa.c
+++ b/libvisual-plugins/plugins/input/alsa/input_alsa.c
@@ -229,6 +229,10 @@ int inp_alsa_upload (VisPluginData *plugin, VisAudio *audio)
 	priv = visual_object_get_private (VISUAL_OBJECT (plugin));
 	visual_log_return_val_if_fail(priv != NULL, -1);
 
+	if (priv->chandle == NULL) {
+		return -1;
+	}
+
 #if 0
 	{	/* DEBUG STUFF, REMOVE IN RELEASE FIXME FIXME XXX TODO WHATEVER */
 		VisBuffer buffer;
--- a/libvisual-plugins/plugins/input/alsa/input_alsa.c
+++ b/libvisual-plugins/plugins/input/alsa/input_alsa.c
@@ -24,6 +24,7 @@
 
 #include <config.h>
 
+#include <math.h>
 #include <stdio.h>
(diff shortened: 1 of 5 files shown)


hartwork:
@kaixiong conflicts fixed and rebased onto latest `0.4.x`.

libvisual-plugins/plugins/input/alsa/input_alsa.c
```diff
@@ -167,14 +168,14 @@ int inp_alsa_init (VisPluginData *plugin)
 
        /* Setup a large buffer */
 
-	tmp = 1000000;
+	tmp = 1000 * 1000 / inp_alsa_var_samplerate * inp_alsa_var_frames_target;  // in micro seconds
```
kaixiong, line 171: This is an integer division as all the operands are integers. The result is likely to be way off.

libvisual-plugins/plugins/input/alsa/input_alsa.c
```diff
@@ -167,14 +168,14 @@ int inp_alsa_init (VisPluginData *plugin)
...
 	if (snd_pcm_hw_params_set_period_time_near(priv->chandle, hwparams, &tmp, &dir) < 0){
 		visual_log(VISUAL_LOG_CRITICAL, _("Error setting period time"));
 		snd_pcm_hw_params_free(hwparams);
 		return(-1);
 	}
 
-	tmp = 1000000*4;
+	tmp = 1000 * 1000 / inp_alsa_var_samplerate * inp_alsa_var_frames_target;  // in micro seconds
```
kaixiong, line 178: This suffers from the integer division problem above.

libvisual-plugins/plugins/input/alsa/input_alsa.c
```diff
@@ -221,14 +222,19 @@ int inp_alsa_upload (VisPluginData *plugin, VisAudio *audio)
 {
 	int16_t data[PCM_BUF_SIZE];
 	alsaPrivate *priv = NULL;
-	int rcnt;
+	const snd_pcm_uframes_t frames_wanted = sizeof(data) / sizeof(data[0]) / inp_alsa_var_channels;
```
kaixiong, line 225: This is another integer division. May be better to use sizeof(data) / (sizeof(data[0]) * inp_alsa_var_channels).

libvisual-plugins/plugins/input/alsa/input_alsa.c
hartwork: @kaixiong I was aware of the integer aspect and took that into consideration when choosing the order of operations. When playing with it locally again, you still seem to have a point. In IPython:

```
In [10]: 1000 * 1000 // 44100 * 256
Out[10]: 5632

In [11]: 1000 * 1000 / 44100.0 * 256
Out[11]: 5804.988662131519

In [12]: 1000 * 1000 * 256 // 44100
Out[12]: 5804

In [13]: 1000 * 1000 * 256 // 44100 + 1
Out[13]: 5805
```

How about `1000 * 1000 * inp_alsa_var_frames_target / inp_alsa_var_samplerate + 1;`? `+1` would be for the fractional part that has been floored away.

kaixiong (commented):
@hartwork, I added some comments regarding the rewritten formulas.

libvisual-plugins/plugins/input/alsa/input_alsa.c
kaixiong: Let's also bracket the product for that extra clarity. +1 is good enough in my view. At worse we allocate 1 more sample/bytes than necessary when the division is perfect.

Is there a chance of overflow with that product? It looks like a pretty large number.

libvisual-plugins/plugins/input/alsa/input_alsa.c
hartwork: @kaixiong we are reversing the two multiplications contained in `int16_t data[PCM_BUF_SIZE]` here and then dividing by the number of channels, literally `(2 * 4096) // 2 // 2` i.e. `(2 * (2 * 2048)) // 2 // 2` i.e. `2048`. As long as `PCM_BUF_SIZE` is an even number, there is no chance for loss here. If it ever is change to something odd, then we lose one sample with or without the braces. Hence my vote for keeping as is, what do you think?

kaixiong: @hartwork, sorry I was a bit too rushed with my original comment. We only support 2 channels at most at the moment, and with that int16_t, we should have divisibility (no loss) in every intermediate step.

libvisual-plugins/plugins/input/alsa/input_alsa.c
hartwork: > @hartwork, let's also bracket the product for that extra clarity.

@kaixiong no objections. Where do you want me to add braces?

> Is there a chance of overflow with that product? It looks like a pretty large number.

On 32bit `signed int tmp` will go up to `2 ** 30 - 1` i.e. `1073741823`. That allows a maximum value of 1073 for `inp_alsa_var_frames_target` before we overflow, or 1024 in power-of-two land:

```
In [4]: (2 ** 30 - 1) / 1_000_000
Out[4]: 1073.741823
```

How about this change:

```diff
--- static const int  inp_alsa_var_frames_target = 256;
+++ static const int  inp_alsa_var_frames_target = 256;  // must be <=1073 to not integer math overflow below
```

libvisual-plugins/plugins/input/alsa/input_alsa.c
hartwork: @kaixiong PS: I can still make that^^ adjustment, if you prefer, no hard feelings about it.

hartwork: @kaixiong only saw your latest comment now. So which version do we use?

libvisual-plugins/plugins/input/alsa/input_alsa.c
kaixiong: @hartwork, it would just look like:
```
tmp = (1000 * 1000 * inp_alsa_var_frames_target) / inp_alsa_var_samplerate + 1
```

Getting nerdsniped by this issue. If we want to get the exact ceiling, I believe we can use this numerical trick:
```
tmp = (1000 * 1000 * inp_alsa_var_frames_target + inp_alsa_var_samplerate - 1) / inp_alsa_var_samplerate
```
Obviously this will reduce the effective range of `inp_alsa_var_samplerate`.

libvisual-plugins/plugins/input/alsa/input_alsa.c
kaixiong: @hartwork, sorry for the confusion. I was happily adding the in-code comments without realising that it didn't need a final submit to show! I think it's okay as is, but if we want to make it a _teeney weeney_ bit nicer, there's a macro called VISUAL_TABLESIZE() for the first part of the equation:

```
const snd_pcm_uframes_t = VISUAL_TABLE_SIZE(data) / inp_alsa_var_channels;
```

I leave it up to you to decide :)

libvisual-plugins/plugins/input/alsa/input_alsa.c
hartwork: @kaixiong maybe we need to collect requirements first. Minimum requirement would be no overflow and to not be off by more than 1. After that, simplicity and precision are in conflict in some sense, at least in pure integer world. With simplicity in mind, `+1` is already not clean and `+ inp_alsa_var_samplerate - 1` is even further from explaining itself, at least quickly. And both of these variants would be off by one if `inp_alsa_var_frames_target` is allowed to change freely. Maybe it's time to go for actual `ceil` and floating point to be clean, precise and self-explaining(?):
```
tmp = (int)ceil((1000.0 * 1000.0 * inp_alsa_var_frames_target) / inp_alsa_var_samplerate);
```
What do you think?

libvisual-plugins/plugins/input/alsa/input_alsa.c
hartwork: @kaixiong, I found `VISUAL_TABLESIZE` now but the "table" part of the name seems more misleading than helpful, since there is not "table" here. My vote for making a copy (we cannot rename in 0.4.x) named `VISUAL_ARRAY_SIZE` for universal use that is more self-explaining, but outside of this pull request fist. I'll leave that part as is for now then.

kaixiong: > My vote for making a copy (we cannot rename in 0.4.x) named `VISUAL_ARRAY_SIZE` for universal use that is more self-explaining, but outside of this pull request fist. I'll leave that part as is for now then.
@hartwork, agreed.

libvisual-plugins/plugins/input/alsa/input_alsa.c
kaixiong: Agreed, we might as well use floats here as it's not performance sensitive. It's clear and very hard to go wrong.

hartwork: @kaixiong okay great, now pushed with a fix.

libvisual-plugins/plugins/input/alsa/input_alsa.c
hartwork: Addressed above at https://github.com/Libvisual/libvisual/pull/154#discussion_r1070397953

kaixiong (approved):
Nice work. Looks like it also fixes a leak with `strdup()` on the ALSA device string.

hartwork:
@kaixiong glad you like it, and thanks for the review! Let me check how much can and needs to be forward-ported to `master`.

PR #159, opened 2023-01-11
libvisual-plugins: Fix mplayer input plugin

hartwork:
Similar to pull request #158 but for `master`.

### DIFF
--- a/libvisual-plugins/plugins/input/mplayer/input_mplayer.c
+++ b/libvisual-plugins/plugins/input/mplayer/input_mplayer.c
@@ -25,6 +25,7 @@
 #include "config.h"
 #include "gettext.h"
 
+#include <limits.h>
 #include <unistd.h>
 #include <fcntl.h>
 #include <string.h>
@@ -52,6 +53,8 @@ typedef struct {
 	int fd;                    /**< file descriptor to mmaped area */
 	char *sharedfile;          /**< shared file name */
 	mplayer_data_t *mmap_area; /**< mmap()'ed area */
+	unsigned long long prev_count; /**< to detect duplicates */
+
 	int loaded;                /**< plugin state */
 } mplayer_priv_t;
 
@@ -157,6 +160,7 @@ static int inp_mplayer_init( VisPluginData *plugin )
 				strerror( errno ) );
 		return FALSE;
 	}
+	priv->prev_count = ULLONG_MAX;
 
 	priv->loaded = TRUE;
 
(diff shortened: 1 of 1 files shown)


kaixiong (commented):
Nice fix. A few points:

- I'm not entirely comfortable with setting an unsigned integral type to `-1` even though the C standard seems to guarantee the wrap around. Perhaps assigning `ULLONG_MAX` (from C99's `<limits.h>`) would be better.
- Do you know if MPlayer is guaranteed to produce a new and complete 'frame' each time we read from the mmap'ed block (assuming that the counter changes)?

hartwork:
> Nice fix. A few points:
> 
> * I'm not entirely comfortable with setting an unsigned integral type to `-1` even though the C standard seems to guarantee the wrap around. Perhaps assigning `ULLONG_MAX` (from C99's `<limits.h>`) would be better.

I don't share the concern (and an explicit cast would be an option) but I have adjust it as suggested.


> * Do you know if MPlayer is guaranteed to produce a new and complete 'frame' each time we read from the mmap'ed block (assuming that the counter changes)?

I believe it's full 512 frames every time. If it's not, the interface has no field of telling the count. The man page says this:

> ```
>        export[=mmapped_file[:nsamples]]
>               Exports the incoming signal to other processes using memory mapping (mmap()).  Memory mapped areas contain a header:
> 
>               int nch                      /*number of channels*/
>               int size                     /*buffer size*/
>               unsigned long long counter   /*Used to keep sync, updated every
>                                              time new data is exported.*/
> 
>               The rest is payload (non-interleaved) 16 bit data.
>                  <mmapped_file>
>                       file to map data to (default: ~/.mplayer/mplayer-af_export)
>                  <nsamples>
>                       number of samples per channel (default: 512)
> 
>               EXAMPLE:
>                  mplayer -af export=/tmp/mplayer-af_export:1024 media.avi
>                       Would export 1024 samples per channel to '/tmp/mplayer-af_export'.
> ```

What strikes me as odd right now is the "non(!)-interleaved" in there :thinking: When testing with https://www.kozco.com/tech/LRMonoPhase4.wav and `mplayer -af export LRMonoPhase4.wav` locally the data seems interleaved rather than non-interleaved to me.

hartwork:
@kaixiong PS: I just "verified" that at least in my local experiments `int size` is the same value (2048) every time.

kaixiong (approved):
Looks good!

PR #153, opened 2023-01-07
[0.4.x] libvisual-plugins: Base rotation on runtime not FPS for actor "lv_gltest"

hartwork:

PR #163, opened 2023-01-13
libvisual-plugins: Base rotation on runtime not FPS for actor "lv_gltest"

hartwork:
Similar to pull request #153 and #164 but for `master`.


kaixiong (approved):
Good to go.

kaixiong:
Wait.. I think the timer needs to be destroyed with visual_timer_free().

hartwork:
> Wait.. I think the timer needs to be destroyed with visual_timer_free().

@kaixiong good catch! Added now and also backported to 0.4.x in pull request #164.

kaixiong (approved):

PR #170, opened 2023-01-15
libvisual-plugins: Replace use of t1lib in DancingParticles with FreeType 2 (closes #53)

### LINKED ISSUE #53: actor_dancingparticles: get rid of t1lib dependency
could be replaced by FreeType

kaixiong:
As title describes.

Rendered text is clipped by a few pixels due to the bounding box estimation method not accounting for anti-aliasing.

This should not be a problem as the text bitmap is only used to calculate some numbers for visualization (without regards for shape or structure).

It's impossible to test this change as it requires song title info, not currently available in LV. I think it also requires the ability to set the right plugin parameters.

However, I have manually verified the change using a separate program utilising the new rendering function:

image


kaixiong:
@hartwork, this is my first attempt at a rebase+force push. Github says there are 4 failing checks but I can't tell what they are.

hartwork:
> @hartwork, this is my first attempt at a rebase+force push. Github says there are 4 failing checks but I can't tell what they are.

@kaixiong one is https://github.com/Libvisual/libvisual/actions/runs/3920882762/jobs/6702767069#step:10:152:

> ```
> [..]/libvisual-plugins/plugins/actor/dancingparticles/gl.cpp:91:16: error: variable 'r' set but not used [-Werror,-Wunused-but-set-variable]
>   static float r=0, SIZE=1;
>                ^
> ```

kaixiong:
> > @hartwork, this is my first attempt at a rebase+force push. Github says there are 4 failing checks but I can't tell what they are.
> 
> @kaixiong one is https://github.com/Libvisual/libvisual/actions/runs/3920882762/jobs/6702767069#step:10:152:
> 
> > ```
> > [..]/libvisual-plugins/plugins/actor/dancingparticles/gl.cpp:91:16: error: variable 'r' set but not used [-Werror,-Wunused-but-set-variable]
> >   static float r=0, SIZE=1;
> >                ^
> > ```

@hartwork That is odd... I recreated the dancingparticles-tf2 branch off master and applied the changes. Then I force-pushed into origin/dancingparticles-tf2. This fix should've been in or master cannot possibly compile.

kaixiong:
@hartwork. Ah never mind. The fix is from the same branch (was never pushed to master) which I accidentally ended up deleting.

kaixiong:
@hartwork, I should also mention that this could be backported to 0.4.x for the next release, with very little source code changes.

kaixiong:
@hartwork and I discussed this PR at length separately and there are a few issues that cropped up that we should document here for anyone who might be watching.

* `Centers` appear to a vector/array that holds particle positions. Dancing Particles does appear to use generate particles from the rasterised text bitmap (where the pixel is non-black).
* The counter variable `numPart` is used to index `Centers` and it is incremented at least once (?) for each pixel in the text bitmap. This means that the text bitmap size does matter. We need to ensure that it never exceeds `ptsNumMax` to prevent buffer overflows.
* `maxscore` is calculated by adding the 8-bit gray value at each pixel of the text bitmap. There is a chance of overflow if the bitmap is too large (too many pixels), especially if LV is built for 32-bit platforms. As `maxscore` is a _signed_ int, an overflow creates undefined behaviour. In the best(?) case, we have a negative number and the compiler does not 'miscompile' this code under aggressive optimizations. But this screws up the calculations that follow later (who expects a max value to be < 0?).

hartwork (commented):
Hi @kaixiong,

I have some news on this topic.  I checked out the first 1,5 FreeType2 tutorials and ended up writing debugging-only code that dumps the old image produced by t1lib as well as the new image producted by FreeType, both in "bad ascii art" style to the terminal.  The debugging code I wrote and related output is in these two new branches:

- [`dancingparticles-t1lib-debug`](https://github.com/Libvisual/libvisual/commits/dancingparticles-t1lib-debug)
- [`dancingparticles-ft2-debug`](https://github.com/Libvisual/libvisual/commits/dancingparticles-ft2-debug)

The [`dancingparticles-ft2-debug` branch](https://github.com/Libvisual/libvisual/commits/dancingparticles-ft2-debug) contains the output of before and after so you can compare them locally:
```bash
cd libvisual-plugins/plugins/actor/dancingparticles
meld t1render.txt ft2render.txt
```

Through creating that debug code I found that:

- The image rendered by t1lib is 103x17 in size, the FreeType2 one is 103x16.
- Function `FT_Set_Pixel_Sizes` does not request pixel sizes in practice despite the name, function `FT_Set_Char_Size` seems to produce better results:
  - `FT_Set_Pixel_Sizes(face, 100, 100)` yields 74x46 for a small "g"
  - `FT_Set_Char_Size(face, 0, 100 * 64, 0, 96 /*dpi*/)` yields 97x62 for a small "g"
- The new code prints `Assertion '!area.empty ()' failed` and crashes with a segfault while converting the image to particle centers; may also be a problem with my debug code in `ft2render.cpp`, not sure yet.
- The old code writes one center pertical too much to the array, see file `t1render.txt` output.
- I found a way to make the actor show the `loadString` results to us as the user would see them, see my change to `etoile.cpp`.
- The `maxscore` calculation is less brittle than previously expected, it takes a (simplified) 3000x3000 picture (`math.sqrt((2**31-1) / 255) --> 2901`), to wrap around.

What do you think?

PS: My plan is to first hear back from you on these findings and the big picture, and to then pick up code review on details again.

Best, Sebastian

kaixiong:
@hartwork 

> Through creating that debug code I found that:
> 
>     * The image rendered by t1lib is 103x17 in size, the FreeType2 one is 103x16.
> 
>     * Function `FT_Set_Pixel_Sizes` does not request pixel sizes in practice despite the name, function `FT_Set_Char_Size` seems to produce better results:
>       
>       * `FT_Set_Pixel_Sizes(face, 100, 100)` yields 74x46 for a small "g"
>       * `FT_Set_Char_Size(face, 0, 100 * 64, 0, 96 /*dpi*/)` yields 97x62 for a small "g"
> 

`FT_Set_Char_Size()` doesn't do anything special here that makes it more accurate. Its size parameters are in _points_, which is defined to be 1/72 inches. Translating that to pixel units give approximately 133.33 (100 * 1/72 * 96).

So the `FT_Set_Char_Size()` call is the equivalent of calling `FT_Set_Pixel_Size(face, 0, 133.33)`, which produces an output that is 133.33% as large.

There are a couple of reasons why the  'g' glyph bitmap is so much shorter than 100px. The most important one is that font height measurements are based on the [Em square](http://designwithfontforge.com/en-US/The_EM_Square.html).

A picture says a thousand words:

image

The baseline is in red. The EM square is the black border around H.

The distance from the baseline to the top of the highest glyph and bottom of the lowest glyph is known as the ascent and descent respectively. A tight fitting bitmap for the text string would have a height of ascent + descent.

>     * The new code prints `Assertion '!area.empty ()' failed` and crashes with a segfault while converting the image to particle centers; may also be a problem with my debug code in `ft2render.cpp`, not sure yet.

This seems unlikely. The assertion is in `LV::Video::create_sub()` and is most likely caused by the `visual_video_blit()` call. Either the glyph bitmap was completely clipped (unsupported) or it's a bug with `LV::Video::create_sub()` (subvideos are kind of broken at the moment).

>     * The old code writes one center pertical too much to the array, see file `t1render.txt` output.
Interesting!

>     * The `maxscore` calculation is less brittle than previously expected, it takes a (simplified) 3000x3000 picture (`math.sqrt((2**31-1) / 255) --> 2901`), to wrap around.

Indeed, I made the same calculation the other day and came to the same conclusion.

I also worked out that the given input character limit of `numPts/50` effectively means that each character glyph may not contain more than 50 non-black pixels on average. We could write a small program to check the pixel counts for the set of characters `loadString()` accepts so that the limit is respected (or we could increase the divisor).
 
> What do you think?

This is going to take a while to properly fix :smile:.

I went through the rendering code again and noticed that the characters are rendered far too low. The baseline is right at the bottom edge of the text bitmap, so all descenders are clipped:

image

After adjusting the y position using the ascender and descender font metrics, I got:

image

We still have pretty serious clipping issues. Since Dancing Particles ignores black pixels, we could give up on finding the correct bounding box (adjusted for antialiasing) and just add ample padding for the glyphs to spill into.

I'm interested in reading the t1lib code but the link from https://www.t1lib.org is FTP and broken.

kaixiong:
@hartwork I've figured out that the clipping issue is entirely due to the distance between the pen position and bottom-left box corner of the first glyph. Translating the glyph positions by (-xMin, -yMin) gives the correct result, even with anti-aliased outlines:

image

So what's left to do is to ensure that Dancing Particles does not overflow the `Centers` array.

hartwork:
@kaixiong I have rebased `dancingparticles-ft2-debug` onto your latest changes now and and synced `ft2render.txt` with actual output. Looks a lot better now :+1: 

>[..] The most important one is that font height measurements are based on the [Em square](http://designwithfontforge.com/en-US/The_EM_Square.html). [..]

You're right, thanks for the explanation! I can confirm now that both…

- `FT_Set_Pixel_Sizes(face, 100, 100)` and
- `FT_Set_Char_Size(face, 0, 100 * 64, 0, 72)`

…get me an EM square size of 100x100 pixels according to:
- `face->size.metrics.x_ppem` and
- `face->size.metrics.y_ppem`.

> > * The new code prints `Assertion '!area.empty ()' failed` and crashes with a segfault while converting the image to particle centers; may also be a problem with my debug code in `ft2render.cpp`, not sure yet.
> 
> This seems unlikely. 

But it was real :smile: After rebasing onto your latest changes, it's gone (but `LV::System::destroy();` crashes with `free(): invalid pointer` now, see `ft2render.cpp`). Not sure yet why or where, could be unrelated.

> I'm interested in reading the t1lib code but the link from https://www.t1lib.org is FTP and broken.

- Code browser: https://fossies.org/dox/t1lib-5.1.2/t1aaset_8c_source.html#l01026
- Download: https://fossies.org/linux/misc/old/t1lib-5.1.2.tar.gz
- Expected SHA256: `821328b5054f7890a0d0cd2f52825270705df3641dbd476d58d17e56ed957b59`

libvisual-plugins/plugins/actor/dancingparticles/font.cpp
```diff
@@ -0,0 +1,229 @@
...
+#include <algorithm>
+#include <limits>
+#include <vector>
+
+#include <ft2build.h>
+#include FT_FREETYPE_H
+#include FT_GLYPH_H
+
+const char *fontPath = DATA_DIR "/font.pfb";
+const int   fontSize = 25; // Does this match the T1lib size parameter?
```
hartwork, line 15: My vote for dropping that comment now, and renaming the variable to `fontSizePoints` or similar to communicate the unit behind the number.

libvisual-plugins/plugins/actor/dancingparticles/font.cpp
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+  error = FT_New_Face(ftLibrary, fontPath, 0, &ftFace);
+  if(error)
+    {
+      visual_log(VISUAL_LOG_WARNING, "Unable to load font (%s).", fontPath);
+      return false;
+    }
+
+  FT_Set_Pixel_Sizes(ftFace, 0, fontSize);
```
hartwork, line 44: No error checking here?

libvisual-plugins/plugins/actor/dancingparticles/font.cpp
```diff
@@ -0,0 +1,229 @@
...
+      textBBox.yMax = std::max(textBBox.yMax, glyphBBox.yMax);
+    }
+
+  if(textBBox.xMin > textBBox.xMax)
+    return nullptr;
+
+  // Rasterize text.
+
+  int textWidth = textBBox.xMax - textBBox.xMin;
+  int textHeight = textBBox.yMax - textBBox.yMin;
```
hartwork, line 193: Should we add 2x `const` here? Just an idea.

libvisual-plugins/plugins/actor/dancingparticles/font.cpp
```diff
@@ -0,0 +1,229 @@
...
+      glyphBBox.yMax += glyph.pos.y;
+
+      textBBox.xMin = std::min(textBBox.xMin, glyphBBox.xMin);
+      textBBox.yMin = std::min(textBBox.yMin, glyphBBox.yMin);
+      textBBox.xMax = std::max(textBBox.xMax, glyphBBox.xMax);
+      textBBox.yMax = std::max(textBBox.yMax, glyphBBox.yMax);
+    }
+
+  if(textBBox.xMin > textBBox.xMax)
+    return nullptr;
```
hartwork, line 188: Do we need to loop `FT_Done_Glyph` here?

libvisual-plugins/plugins/actor/dancingparticles/font.cpp
kaixiong: Good idea. Const should be applied wherever possible.

libvisual-plugins/plugins/actor/dancingparticles/font.cpp
kaixiong: Yes, we do. I will work it into the Glyph class destructor to avoid manual cleanups.

libvisual-plugins/plugins/actor/dancingparticles/font.cpp
kaixiong: I agree with dropping the comment. It should be renamed `fontSizePixels` though.

hartwork: @kaixiong right! Maybe a new comment that these are about the EM square would help the next developer, just an idea.

hartwork (changes requested):
@kaixiong with the latest changes (commit 634c09a2fbf92f208dd547b398a1240916a8b942), I'm getting crashes with ASan and -O1 now at `font.cpp:235`:

```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==19345==ERROR: AddressSanitizer: SEGV on unknown address 0x000075800002 (pc 0x7fd7a8494643 bp 0x607000004540 sp 0x7ffce8219820 T0)
==19345==The signal is caused by a READ memory access.
    #0 0x7fd7a8494643 in FT_Glyph_To_Bitmap (/usr/lib64/libfreetype.so.6+0x1e643)
    #1 0x560990262282 in rasteriseText(FT_FaceRec_*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) [..]/libvisual-plugins/plugins/actor/dancingparticles/font.cpp:235
    #2 0x560990262b77 in loadString(char const*) [..]/libvisual-plugins/plugins/actor/dancingparticles/font.cpp:96
    #3 0x560990260ae2 in main [..]/libvisual-plugins/plugins/actor/dancingparticles/ft2render.cpp:20
    #4 0x7fd7a7e8a209  (/lib64/libc.so.6+0x2a209)
    #5 0x7fd7a7e8a2bb in __libc_start_main (/lib64/libc.so.6+0x2a2bb)
    #6 0x560990260840 in _start ([..]/libvisual-plugins/build_cmake/plugins/actor/dancingparticles/ft2render+0x7840)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV (/usr/lib64/libfreetype.so.6+0x1e643) in FT_Glyph_To_Bitmap
==19345==ABORTING
```

kaixiong:
@hartwork, thanks for the report. The issue is most likely due to use after free. Sorry I should've spent more time verifying this before committing.

kaixiong:
@hartwork, could you verify with the latest commit [856ab82](https://github.com/Libvisual/libvisual/pull/170/commits/856ab82ec60f243c94260fa64244a658866be03b). I ran it against ASan and the only errors I see are related to subvideos (which we already know are problematic).

hartwork:
> @hartwork, could you verify with the latest commit [856ab82](https://github.com/Libvisual/libvisual/pull/170/commits/856ab82ec60f243c94260fa64244a658866be03b). I ran it against ASan and the only errors I see are related to subvideos (which we already know are problematic).

@kaixiong I'm getting a different error now:
```
=================================================================
==25461==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x62a000005fc0 at pc 0x5575ddc53b44 bp 0x7ffdc044c5a0 sp 0x7ffdc044c590
WRITE of size 4 at 0x62a000005fc0 thread T0
    #0 0x5575ddc53b43 in FloatPoint::operator=(FloatPoint) [..]/libvisual-plugins/plugins/actor/dancingparticles/floatpoint.h:87
    #1 0x5575ddc53b43 in loadString(char const*) [..]/libvisual-plugins/plugins/actor/dancingparticles/font.cpp:132
    #2 0x5575ddc51b22 in main [..]/libvisual-plugins/plugins/actor/dancingparticles/ft2render.cpp:20
    #3 0x7f0be040b209  (/lib64/libc.so.6+0x2a209)
    #4 0x7f0be040b2bb in __libc_start_main (/lib64/libc.so.6+0x2a2bb)
    #5 0x5575ddc51880 in _start ([..]/libvisual-plugins/build_cmake/plugins/actor/dancingparticles/ft2render+0x8880)

0x62a000005fc0 is located 0 bytes to the right of 24000-byte region [0x62a000000200,0x62a000005fc0)
allocated by thread T0 here:
    #0 0x7f0be0be1937 in __interceptor_malloc /var/tmp/portage/sys-devel/gcc-11.3.1_p20221223/work/gcc-11-20221223/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x5575ddc51ae0 in main [..]/libvisual-plugins/plugins/actor/dancingparticles/ft2render.cpp:15
    #2 0x7f0be040b209  (/lib64/libc.so.6+0x2a209)

SUMMARY: AddressSanitizer: heap-buffer-overflow [..]/libvisual-plugins/plugins/actor/dancingparticles/floatpoint.h:87 in FloatPoint::operator=(FloatPoint)
Shadow bytes around the buggy address:
  0x0c547fff8ba0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c547fff8bb0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c547fff8bc0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c547fff8bd0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c547fff8be0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
=>0x0c547fff8bf0: 00 00 00 00 00 00 00 00[fa]fa fa fa fa fa fa fa
  0x0c547fff8c00: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c547fff8c10: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c547fff8c20: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c547fff8c30: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c547fff8c40: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
  Shadow gap:              cc
==25461==ABORTING
```

kaixiong:
@hartwork That looks like an out of bounds access when copying/storing a `FloatPoint` value into `Centers`. It's something we anticipated.

What is surprising to me is that a short string like 'Moeders' is already exceeds the array limit.

So we'll have to focus on figuring out the following:
1. Correctly sizing `Centers` to accommodate the rasterised text
2. Reducing the font size or truncating the input text at the right length to reduce the non-black pixel count

hartwork:
> @hartwork That looks like an out of bounds access when copying/storing a `FloatPoint` value into `Centers`. It's something we anticipated.

@kaixiong you're right! I didn't see it because I was pumping the output into file `ft2render.txt`…


> What is surprising to me is that a short string like 'Moeders' already exceeds the array limit.

I believe it tries to produce 2001 entries points of data independent of string length.


> So we'll have to focus on figuring out the following:
> 
>     1. Correctly sizing `Centers` to accommodate the rasterised text
> 
>     2. Reducing the font size or truncating the input text at the right length to reduce the non-black pixel count

I'm not sure if that's the right direction, if it's 2001 entries for all inputs.
(Though I am also not yet sure I understand the original `while(lastscore < scoreToGo  && curscore >= scoreToGo)` idea yet, really.)

PS: I think I haven't mentioned this yet, but since neither `c` nor `curscore` goes into the right side of `Centers[numPart++]= ...`, we could in theory tell FreeType to render in monochrome mode.

hartwork:
@kaixiong I am starting to think that maybe we're trying to accomplish too much at once here when the original mission was to become independent of t1lib. How about we add out-of-bounds-protection a la…
```c
while(lastscore < scoreToGo && curscore >= scoreToGo)
  {
    if (numPart >= ptsNum)  // <-- this
      {                     // <-- part
        return;             // <-- right
      }                     // <-- here :)
    Centers[numPart++] = [..];
    [..]
  }
```
…to the loop and go back to the drawing board with regard to particles location algorithm, some other time?

kaixiong:
@hartwork I figured out the algorithm of the text-to-particle generator.

Essentially the generator tries to allocate a total of `ptsNum` particles to all non-black pixels in the text string bitmap. The particle allocation to each pixel is proportionate to its brightness.

It mostly looks correct, with the exception of two problems:
- There is an off-by-one array overflow. It can be fixed by adjusting the `scoreToGo` formula to use `numCenters-1`.
- The algorithm fails miserably when `numCenters > maxScore`, producing only 1 particle. `scoreToGo` goes gets stuck at 1 or goes to 0 because of the use of integer division. This shouldn't be hard to fix using floating point arithmetic, or using an incremental integer error algorithm (Bresenham's line algorithm is a classic example).

kaixiong:
@hartwork Would appreciate if you can help test the latest commit with the off-by-one error fix. I've only tested it myself on a standalone, minimalist excerpt of  `loadString()`. I have some link errors building LV and LV plugins with ASAN even with what looks like the right linker flags. The script you sent me has the same problem.

hartwork:
> @hartwork I figured out the algorithm of the text-to-particle generator.
> 
> Essentially the generator tries to allocate a total of `ptsNum` particles to all non-black pixels in the text string bitmap. The particle allocation to each pixel is proportionate to its brightness.

@kaixiong I'm not sure I understand.

> @hartwork Would appreciate if you can help test the latest commit with the off-by-one error fix.

I updated branch `dancingparticles-ft2-debug` with the latest output. Numbers changed and there no longer is a write to index 2000. (When running `lv-tool acter dancingparticles` I get a crash in `changep()` at `parameters.cpp:290` with ASan after a second or so of runtime on the `-debug` branch, but it seems unrelated.)

> I have some link errors building LV and LV plugins with ASAN even with what looks like the right linker flags. The script you sent me has the same problem.

Let's debug that here or off-GitHub and find the cause. ASan is too essential to not be working :smiley:

kaixiong:
@hartwork

> @kaixiong I'm not sure I understand.

The function first tallies up the gray level of each pixel in the bitmap and stores that in `maxScore`.

Next, it looks at every pixel. Let's say the pixel's gray level is `c`. It then generates roughly `(c * ptsNum) / maxScore ` particles for that pixel. The brighter the pixel is, the more particles will be created for it. This is what the inner while loop does.

Let's say `maxScore` was found to be 10000, `ptsNum` is 2000.

|Gray Level| # Particles | Calculation |
|---------------|-----------------|------------------|
|255 |51| (255 * 2000.0) / 10000 = 51|
|128 |25 or 26| (128 * 2000.0) / 10000 = 25.6|
| 64 | 12 or 13 | (64 * 2000.0) / 10000 = 12.8|
| 0    | 0 | (0 * 2000.0) / 10000 = 0|

> Let's debug that here or off-GitHub and find the cause. ASan is too essential to not be working :smiley: 

Yeah, agreed.

hartwork:
@kaixiong thanks for the detailed explanation! At first expression `(c * ptsNum) / maxScore` looked the wrong way around, but by now I can confirm. Here's how I got there if curious, you probably did something similar:

```
(numPart * maxscore) / (numCenters - 1) ~= score 
  <->                                            | assuming numCenters != 1
numPart * maxscore ~= score * (numCenters - 1)
  <->                                            | assuming maxscore != 0
numPart ~= score * (numCenters-1) / maxscore
  ->
particle_count(c) := c * ptsNum / maxscore
```

> > Let's debug that here or off-GitHub and find the cause. ASan is too essential to not be working smiley
> 
> Yeah, agreed.

Alright, what's the link error?

kaixiong:
@hartwork, I rewrote the algorithm using clearer formulas. Let me know what you think.

The generated y-coordinates is slightly different at places, (up to 4.0 every few pixels I believe), due to the different rounding. The rewrite is probably more accurate however. In any case, it's not noticeable to me in the rendered output.

hartwork:
@kaixiong it seems the new implementation also changes how many particles certain greys get if I'm reading the diff at 7cb000b2c20d07199b8d4692a4c89d4465fb2fb4 correctly, unsure if for good or for bad. On a more detail level, I'm not sure about the term "energy" used there and I have a feeling that we may get off-by-ones for `assert(particleCount == ptsNum);`. For the `0.0` I'm wondering if should be `0.0f` if its siblings are cast to float. Also wondering if we should resolve `numCenters` and `numCenters2` when at a re-write. `allocCount` could use `const`.

kaixiong:
@hartwork

> @kaixiong it seems the new implementation also changes how many particles certain greys get if I'm reading the diff at [7cb000b](https://github.com/Libvisual/libvisual/commit/7cb000b2c20d07199b8d4692a4c89d4465fb2fb4) correctly, unsure if for good or for bad.

Yes that is to be expected. It is down to differences in rounding. Note that the rounding in the original algorithm is in the inner loop, with how `scoreToGo` is calculated. I have no reason to believe the difference is significant or problematic.

As mentioned earlier, the original algorithm has problems coping with `numCenters > maxscore` with some inputs (all?). The rewritten version shrugs at that. This condition is unlikely to happen given the current parameters.

The bigger issue I have with the original algorithm is that it is not as clear, and it's harder to show that it terminates with exactly `ptsNum` particles. It's easy for the rewritten version, as I will show later.

> On a more detail level, I'm not sure about the term "energy" used there and I have a feeling that we may get off-by-ones for `assert(particleCount == ptsNum);`. 

I used the term energy because it's a gray level, which I likened to brightness / light energy of a pixel. I thought it was at least better than `score` because a summation of energy makes sense. Each particle could then be thought of as a quantum.

 `assert(particleCount == ptsNum);` never fails, and it's easy to prove for the rewritten algorithm:

The key formula to look at is:
```c++
int allocCount = (energySoFar * numCenters) / totalEnergy - particleCount;
```
Let's break up the equation into two to make it a bit simpler to state:
```c++
int particleCountToReach = (energySoFar * numCenters) / totalEnergy
int allocCount = particleCountToReach - particleCount;
```
`particleCountToReach` reaches maximum at the last pixel. With the gray level of the last pixel included in `energySoFar`, `energySoFar` must be equal to `totalEnergy` at this point. So  `particleCountToReach` is equal to `numCenters`.

The `for` loop that follows _always_ terminates with `particleCount == particleCountToReach`. After the last pixel is processed, `particleCount` must equal to `numCenters`.

Whether assertion is always true is trickier to prove for the original algorithm.

There is a chance for overflow though. That can be solved switching to floating:

```c++
int allocCount = int(float(energySoFar)/totalEnergy * numCenters) - particleCount;
```

> For the `0.0` I'm wondering if should be `0.0f` if its siblings are cast to float.

Could be changed for stylistic improvement.

> Also wondering if we should resolve `numCenters` and `numCenters2` when at a re-write. `allocCount` could use `const`.

That is way out of scope in my view. I already take issue with how `Centers` is allocated and managed, so there will be no end if we follow through.

hartwork:
@kaixiong I follow your explanation on termination, thanks.

Regarding resolution `numCenters` and `numCenters2` I was not aware or forgot that these are used elsewhere — you're right, it's out of scope.

Regarding term "energy", "lightness" seems to be a better match in terminology if thinking grayscale; "opacity" would be better term if we think FreeType alpha channel, which is what the docs say we get from their API calls: An alpha channel. I favor "lightness" for simplicity, "energy" just seems too far off.

How would you like to continue?

hartwork:
@kaixiong PS: My vote for `double` on the overflow topic. I think we risk losing precision if `float` has no more bits than `int`.

kaixiong:
@hartwork

> Regarding term "energy", "lightness" seems to be a better match in terminology if thinking grayscale; "opacity" would be better term if we think FreeType alpha channel, which is what the docs say we get from their API calls: an alpha channel. I favor "lightness" for simplicity, "energy" just seems too far off.

Alpha as used by FreeType's documentation is not actually opacity but _coverage_. The value at each pixel is a measure of how much of its area is in the glyph outline (0% to 100%), scaled to 0-255. This is the  'proper' and original interpretation of alpha by its inventors i.e. Porter and Duff.

That paper, if I recall correctly, discussed image compositing with the addition of a coverage layer called alpha and introduced algebraic operators to describe the various types of composition (beautiful stuff!). It also noted that coverage be reinterpreted as opacity i.e. how much light is blocked at each pixel. In other words, these two notions are interchangeable at the subpixel level.

I've read a fair amount of computer graphics literature. 'Lightness' is not really a term I've come across. 'Brightness' is common but vague as it are often confused with subjective perceptions of colour. 'Energy' and associated terms like radiance, flux, etc. come from physically-based rendering, as that involves the physics of light transport. In digital imaging, the term 'luminance' (or 'luma') is common, which ultimately is based on energy.

For me, the pick is between either luma/luminance or energy. Or we could stick to 'c' and leave it at that.

For what it's worth, I've ran some tests to compare the difference between the two algorithms. I calculated the expected particle account using `(c/c_total) * total_particle_count` and compared the integer allocated count against it. 

In general, the two algorithms looks they're roughly tied on RMS error (I haven't gone so far as to check which method had the lesser RMS error most of the time!). But there are specific inputs where the original produces very large errors. Here's an example.

image

This was measured on an manually drawn circle on a 8x4 grid, 1000 particles in total. Black pixels were not included in the tabulation.

hartwork:
@kaixiong I took the term lightness from [HSL](https://en.wikipedia.org/wiki/HSL_and_HSV) — hue, saturation, lightness. I'd be okay with brightness, maybe luminance or intensity also work. I consider "c" better than "energy", I'd vote for using a term that a reader will understand that does not have a degree in computer graphics or physics, that's my main motivation. Language can be excluding without need to; let's keep the barrier to entry at a human level, we're already raising the the barrier by using C++20 by quite a bit.

Nice table and analysis, looks like the results got better :+1: 

PS: Thanks for going for more `const` and `static_cast` in cd59f3ab8587f917da032a1468bea40b8ab6eea1.

kaixiong:
> @kaixiong I took the term lightness from [HSL](https://en.wikipedia.org/wiki/HSL_and_HSV) — hue, saturation, lightness.

@hartwork Ah! HSL and HSV are perceptual colour spaces though.

> I'd be okay with brightness, maybe luminance or intensity also work. I consider "c" better than "energy", I'd vote for using a term that a reader will understand that does not have a degree in computer graphics or physics, that's my main motivation. Language can be excluding without need to; let's keep the barrier to entry at a human level, we're already raising the the barrier by using C++20 by quite a bit.

I'm fine with 'c'. It somehow feels right, even though it's actually more abstract than the score/energy analogies that's been cooked up.

> PS: Thanks for going for more `const` and `static_cast` in [cd59f3a](https://github.com/Libvisual/libvisual/commit/cd59f3ab8587f917da032a1468bea40b8ab6eea1).

No problem. I've done enough programming in Rust and other immutable-by-default languages (OCaml, Haskell) to appreciate the value of this approach :smile:.

kaixiong:
@hartwork, I've renamed 'energy' to 'C'.

You mentioned earlier than the particle count calculation should use a `double`. After calculating the errors involved and doing some testing, I've concluded that this is not necessary, so long as the text string shorter or equal to 105 characters, and there are at most 5250 particles. This is more than double the current limit and uses very conservative assumptions about the font and input string.

The rough reason why floats work is that the mantissa can hold up to 23 bits, so we will not lose any accuracy so long as `totalC =< 2^24 - 1`. Suppose that each rasterised character/glyph is a _solid white_ block of 25x25 at font height 25. This works out to be around `floor((2^24 - 1)/25/25/255) = 105` characters. Since dancingparticles clips the text to (numPts/50) characters, so we have a particle limit of `105*50=5250`.

hartwork:
@kaixiong thanks for the change and the explanation. I think that what works today and what is robust towards future changes by other authors are slightly different topics (and a conscious decision). It's good to know that we're okay today, but if a future change can break things easily versus has a hard time breaking it, that's a big difference to me (in particular while this very code does not have an automated test to catch regressions). On the math side, I do not yet understand how 23 bits float mantissa could be lossless and better than 32 bits int. What am I missing?

kaixiong:
@hartwork 

> @kaixiong thanks for the change and the explanation. I think that what works today and what is robust towards future changes by other authors are slightly different topics (and a conscious decision). It's good to know that we're okay today, but if a future change can break things easily versus has a hard time breaking it, that's a big difference to me (in particular while this very code does not have an automated test to catch regressions).

I agree. I think it's safe in this case because the margin for mistakes is pretty large.

> On the math side, I do not yet understand how 23 bits float mantissa could be lossless and better than 32 bits int. What am I missing?

It seems odd at first I agree. The short answer is that, for practical parameters, we have overflow issues with integer arithmetic. With floating point arithmetic, we get rounding issues that never gets large enough to affect the result.

Here's the formula we're attempting to calculate:

`(c/c_total) * total_particle_count`

Under integer arithmetic, this will have to be rewritten as `(c * total_particle_count) / c_total` for there not to be precision loss. Why? `c` is smaller than `c_total `until the last iteration where they're equal. The integer rounding of `c/c_total` until then would throw away the _entire_ fractional part. Performing the multiplication first avoids this problem but introduces a potential overflow. If `c_total` is around 2^24, then a `total_particle_count` of 2^8 (only 256) would send it to 2^32, over 2^31-1. The parameters we could use are hard-capped by this on 32-bit platforms.

_Aside: In x86 assembly (and perhaps others), the product of two 32-bit operands would be held in two 32-bit registers (edx:eax) which you can then divide back down without issue, so long as the final result is < 2^32 (fits within eax). We don't have this guarantee in C/C++ as far as I know._

Floats have a far larger range, so this type of overflow simply does not occur for any practical parameters. What we have to contend with are rounding errors due to the 23-bit mantissa. Assuming that all of the variables involved (`c`, `c_total`, `total_particle_count`) are < 2^24, the only rounding errors that can arise are due to multiplication and division. IEEE 754 guarantees that multiplications/divisions are carried out as though infinite precision is available, then rounded down to fit the 23-bit mantissa, giving a maximum relative error of around 2^-23 for each step*, far too small to affect the final result. `total_particle_count` would have to be very close to 2^24 for the rounding error to show up, and even then it'd be off by +/- 1 at most.

(* Does not necessarily accumulate.)

hartwork:
@kaixiong thanks for elaborating. I can confirm that it needs large numbers in that area, the smallest one where I could reproduce a difference in practice was $2^{25}-1$. I don't really like the off by one chance personally, especially if it works against us in the current code — we'd produce one item too many — I would vote for either using `double` or for adding assertions that the value is below a certain constant so we get a clean self-explaining crash (for builds without macro `NDEBUG` defined) at least. What do you think?

My code was this:

```console
# gcc main.c && ./a.out
Via float:  33554432
Via double: 33554431
Difference: 1

# cat main.c 
#include <stdio.h>

int main() {
  const unsigned int accumulatedC = (1 << 25) - 1;
  const unsigned int totalC = accumulatedC;
  const unsigned int numCenters = accumulatedC;

  const int countToReachViaFloat =
      (int)((float)(accumulatedC) / totalC * numCenters);
  const int countToReachViaDouble =
      (int)((double)(accumulatedC) / totalC * numCenters);

  printf("Via float:  %u\n", countToReachViaFloat);
  printf("Via double: %u\n", countToReachViaDouble);
  printf("Difference: %d\n",
         (int)countToReachViaFloat - (int)countToReachViaDouble);

  return 0;
}
```

kaixiong:
@hartwork It's funny how we're getting nerd-sniped by this plugin :laughing:

Do take note that the largest non-power-of-2 integer representable in single-precision FP is 2^24 - 1, so all bets are off if you go higher (which your test code did).

Interestingly enough, Clang warns about this (but not GCC) if the `uint32_t` to `float` conversion is done implicitly:
```
warning: implicit conversion from 'const uint32_t' (aka 'const unsigned int') to 'float' changes value from 16777217 to 16777216 [-Wimplicit-const-int-float-conversion]
```
Anyway, there is no chance of buffer overflow so long as the above is held since `((cTotal-x) / cTotal) <= cTotal / cTotal = 1`

I'm okay with adding either the assertion or converting it to double. I just thought it was interesting enough to figure out why float actually works just fine.

hartwork:
> @hartwork It's funny how we're getting nerd-sniped by this plugin laughing

@kaixiong :smile: 

> Interestingly enough, Clang warns about this (but not GCC) if the `uint32_t` to `float` conversion is done implicitly:
> 
> ```
> warning: implicit conversion from 'const uint32_t' (aka 'const unsigned int') to 'float' changes value from 16777217 to 16777216 [-Wimplicit-const-int-float-conversion]
> ```

Clang 15 started to do that, I love it. We have 39b2d05d65ec47d0310bc427071b24f9c523689b with related fixes on `master`.


> Anyway, there is no chance of buffer overflow so long as the above is held since `((cTotal-x) / cTotal) <= cTotal / cTotal = 1`
> 
> I'm okay with adding either the assertion or converting it to double. I just thought it was interesting enough to figure out why float actually works just fine.

`double` sounds perfect. Yes please! :+1:

kaixiong:
@hartwork 
> `double` sounds perfect. Yes please! +1

It's done.

hartwork:
@kaixiong thanks! :+1:

kaixiong:
@hartwork, are we good to merge then? I think @stuaxo's going gray waiting for this patch to land.

stuaxo:
Heh, don't rush anything on my behalf t :)

hartwork:
> @hartwork, are we good to merge then?

@kaixiong there is one unresolved comment about potentially incomplete cleanup at https://github.com/Libvisual/libvisual/pull/170/files#r1070730416 . What do you think?

Other than that, while I could ask more questions about the ft2 rendering code to fully understand it myself, for the moment I'm good with you understanding it to the last detail and believe that this pull request is a great move in the right direction with "proven" results and solves an important problem, so I'd be happy if we merge this once we are sure where that cleanup is complete (the comment I linked above).

kaixiong:
Oh yes, I should mention that the recently released FreeType 2.13.0 has a Qt-based font viewer. Could come in handy for troubleshooting. Here is a screenshot I grabbed from their [website](https://freetype.org/):

image

hartwork:
> > @kaixiong there is one unresolved comment about potentially incomplete cleanup at https://github.com/Libvisual/libvisual/pull/170/files#r1070730416 . What do you think?
> 
> @hartwork The clean up code is okay as far as I can tell. `FT_Done_FreeType` is guaranteed to release all resources. The [documentation](https://freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_done_freetype) is clear on this:
> 
> _Destroy a given FreeType library object and all of its children, including resources, drivers, faces, sizes, etc._

@kaixiong I see! Would you mind adding something like `ftFace = nullptr;  // already freed by FT_Done_FreeType above` to our code? It think we'll save future readers some time and question marks.

hartwork:
> Oh yes, I should mention that the recently released FreeType 2.13.0 has a Qt-based font viewer. Could come in handy for troubleshooting. Here is a screenshot I grabbed from their [website](https://freetype.org/):

@kaixiong very nice! Super recent indeed, only released yesterday, Arch and Alpine [already have it](https://repology.org/project/freetype/history).

hartwork (approved):

kaixiong:
> @kaixiong very nice! Super recent indeed, only released yesterday, Arch and Alpine [already have it](https://repology.org/project/freetype/history).

@hartwork Good to know but I'm still not going for rolling release distros :wink:

PR #173, opened 2023-01-15
libvisual-plugins: Update actor_gdkpixbuf to build with GTK 4.x.

kaixiong:
The GTK+ 2.0 requirement for actor_gdkpixbuf is a bit dated, so I have updated it to build with GTK 4.x

The `GdkPixbuf` methods used i.e. `new_file_from_file`, `scale_simple` appears unchanged since 2.x. This change required no source code update.

I'm not sure about the deployment status of GTK 4 compared to GTK 3 in the Linux landscape at the moment. On Pop_OS! at least, GTK 4 has been available as part of the default installation for years due to GNOME Shell being the main DE.

We could backport this change to the 0.4.x release. The GTK requirement can be dropped to 3.x without source code updates (if so desired).


hartwork (commented):
Hi @kaixiong 

> The GTK+ 2.0 requirement for actor_gdkpixbuf is a bit dated, so I have updated it to build with GTK 4.x

I see! With https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=967589 in mind, Debian will probably like somethinglike this this (in particular a later backport to 0.4.x).


> The `GdkPixbuf` methods used i.e. `new_file_from_file`, `scale_simple` appears unchanged since 2.x. This change required no source code update.

Cool!


> I'm not sure about the deployment status of GTK 4 compared to GTK 3 in the Linux landscape at the moment. On Pop_OS! at least, GTK 4 has been available as part of the default installation for years due to GNOME Shell being the main DE.

Let me braindump here:
- Ubuntu >=22.04 has GTK 4 — https://packages.ubuntu.com/jammy/libgtk-4-1
- Debian stable/bullseye does not have GTK 4 but they are in "freeze" to the next stable with GTK 4 present, to my understanding — https://packages.debian.org/bookworm/libgtk-4-1
- (I have GTK 4.8.3 on Gentoo)
- Repology is a great resource for questions like that but the case of GTK is not the easiest — https://repology.org/project/gtk/information

> We could backport this change to the 0.4.x release. The GTK requirement can be dropped to 3.x without source code updates (if so desired).

If you're okay with GTK 3, my vote for 3 now and 4 later, to solve things for Debian stable.

What do you think?

kaixiong:
> If you're okay with GTK 3, my vote for 3 now and 4 later, to solve things for Debian stable.
> 
> What do you think?

@hartwork, yes let's go for it.

Should I create a new pull request to replace this one? Dropping the requirement to GTK3 involves replacing the entirety of the changeset. This branch has the 'gtk4' in its name too.

PR #178, opened 2023-01-17
lv-tool: Upgrade default actor plugin to "lv_gltest" if available

hartwork:
This is a foward port of 0.4.x commit e4a6cf5b60771b8cffa90834c0dd967b006bd3b1 from pull request #150 to `master`.

### DIFF
--- a/libvisual/libvisual/lv_actor.h
+++ b/libvisual/libvisual/lv_actor.h
@@ -53,6 +53,15 @@ namespace LV
   public:
 
       /**
+       * Determines if an actor plugin by the given name is available.
+       *
+       * @param name Name of actor plugin to check for
+       *
+       * @return True if an actor plugin by that name is available, else false
+       */
+      static bool available (std::string const& name);
+
+      /**
        * Creates a new Actor with a plugin of a given name.
        *
        * @see realize()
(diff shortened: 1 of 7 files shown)


libvisual/libvisual/lv_actor.h
```diff
@@ -52,6 +52,15 @@ namespace LV
...
+       *
+       * @param name Name of plugin to load
+       *
+       * @return True if an actor by that name is available, else false
+       */
+      static bool available (std::string const& name);
+
       /**
        * Creates a new Actor with a plugin of a given name.
        *
```
kaixiong, line 66: Having this is a good idea. Can we move this into a separate patch and add the method to `LV::Morph` and `LV::Input` as well?

kaixiong (changes requested):

libvisual/libvisual/lv_actor.h
hartwork: @kaixiong done :+1:

kaixiong (changes requested):
@hartwork, thanks. Can you insert a space between the `available()` and its argument list, like this:
```cpp
if (!available (name)) {
```

hartwork:
> @hartwork, thanks. Can you insert a space between the `available()` and its argument list, like this:
> 
> ```c++
> if (!available (name)) {
> ```

@kaixiong done

kaixiong (approved):
@hartwork, all good to go.

PR #182, opened 2023-01-18
[0.4.x] lv-tool: Fix broken --actor argument

hartwork:
0.4.x backport of pull request #181

### DIFF
--- a/libvisual/po/de.po
+++ b/libvisual/po/de.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: libvisual 0.4.0-2.2\n"
 "Report-Msgid-Bugs-To: https://github.com/Libvisual/libvisual/issues/\n"
-"POT-Creation-Date: 2023-01-15 01:44+0100\n"
+"POT-Creation-Date: 2023-01-19 17:24+0100\n"
 "PO-Revision-Date: 2017-08-10 18:46+GMT\n"
 "Last-Translator: Chris Leick <c.leick@vollbio.de>\n"
 "Language-Team: German <debian-l10n-german@lists.debian.org>\n"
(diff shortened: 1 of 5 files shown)


kaixiong (commented):
Is moving the plugin availability check before `parse_args()` runs a good idea on the whole?

There's a chance of a buggy plugin that crashes lv-tool when its info() function is called on loading. When this happens, the user would not be able to access any help options.

On the other hand, it would more reliably produce a working default actor to use.

Sorry to only bring this up now. This escaped my review of the same patch for master.

hartwork:
> @hartwork Is moving the plugin availability check before `parse_args()` runs a good idea on the whole?
> 
> There's a chance of a buggy plugin that crashes lv-tool when its info() function is called on loading. When this happens, the user would not be able to access any help options.
> 
> On the other hand, it would more reliably produce a working default actor to use.
> 
> Sorry to only bring this up now. This escaped my review of the same patch for master.

@kaixiong the good news is that the only plugin that can crash the argument parsing then is lv_gltest where we have control. An alternative could be to leave the actor at empty string initially, and apply defaults only if still unset after argument parsing. Should I demo the latter approach in place, what do you think?

kaixiong:
> @kaixiong the good news is that the only plugin that can crash the argument parsing then is lv_gltest where we have control. An alternative could be to leave the actor at empty string initially, and apply defaults only if still unset after argument parsing. Should I demo the latter approach in place, what do you think?

Yeah let's try that. This would be safer with LV's current way of retrieving metadata by calling an external C/C++ function in the same address space.

hartwork:
> Yeah let's try that. This would be safer with LV's current way of retrieving metadata by calling an external C/C++ function in the same address space.

@kaixiong pushed

hartwork:
> Sorry to only bring this up now. This escaped my review of the same patch for master.

@kaixiong PS: No worries, I can forward-port this to `master` once this is merged.

kaixiong (approved):
Thanks for the change.

PR #184, opened 2023-01-18
[0.4.x] libvisual: Fix ringbuffer infinite loop

hartwork:
One way to see the unfixed behavior in action is to run
1. Fire up `mplayer -af export <some_media_file>`.
2. Run `lv-tool --input mplayer`.
3. Stop mplayer playback or close mplayer.
4. Wait for lv-tool to appear frozen while running in circles in `visual_ringbuffer_get_data_offset`.

### DIFF
--- a/libvisual/libvisual/lv_ringbuffer.c
+++ b/libvisual/libvisual/lv_ringbuffer.c
@@ -416,8 +416,13 @@ int visual_ringbuffer_get_data_from_end (VisRingBuffer *ringbuffer, VisBuffer *d
 	int totalsize = visual_ringbuffer_get_size (ringbuffer);
 	int offset = totalsize - nbytes;
 
-	if ((nbytes / totalsize) > 0)
+	if ((totalsize != 0) && (nbytes > totalsize)) {
 		offset = totalsize - (nbytes % totalsize);
+	}
+
+	if (offset < 0) {
+	    offset = 0;
+	}
 
 	return visual_ringbuffer_get_data_offset (ringbuffer, data, offset, nbytes);
 }
--- a/libvisual/libvisual/lv_ringbuffer.c
+++ b/libvisual/libvisual/lv_ringbuffer.c
@@ -269,7 +269,7 @@ int visual_ringbuffer_get_data_offset (VisRingBuffer *ringbuffer, VisBuffer *dat
 	curposition = buffercorr;
 
 	/* Buffer fixated with partial segment, request the other segments */
-	while (curposition < nbytes) {
+	{
 		int lindex = 0;
 		le = NULL;
 
@@ -326,20 +326,16 @@ int visual_ringbuffer_get_data_offset (VisRingBuffer *ringbuffer, VisBuffer *dat
 			if (curposition == nbytes)
 				return VISUAL_OK;
 		}
-
-		if (entry == NULL) {
-		    // Silence the remaining bytes that we could not fill with
-		    // actual audio
-		    const size_t bytes_to_silence = nbytes - curposition;
-		    void * const target = visual_buffer_get_data (data) + curposition;
-		    memset(target, 0, bytes_to_silence);
-		    return -VISUAL_ERROR_IMPOSSIBLE;
-		}
-
-		startat = 0;
 	}
 
-	return VISUAL_OK;
+	assert (entry == NULL);
+
+	// Silence the remaining bytes that we could not fill with
+	// actual audio
+	const size_t bytes_to_silence = nbytes - curposition;
+	void * const target = visual_buffer_get_data (data) + curposition;
+	visual_mem_set (target, 0, bytes_to_silence);
+	return -VISUAL_ERROR_IMPOSSIBLE;
 }
 
 static int fixate_with_partial_data_request (VisRingBuffer *ringbuffer, VisBuffer *data, int offset, int nbytes,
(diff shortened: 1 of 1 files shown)


libvisual/libvisual/lv_ringbuffer.c
```diff
@@ -416,8 +426,13 @@ int visual_ringbuffer_get_data_from_end (VisRingBuffer *ringbuffer, VisBuffer *d
 	int totalsize = visual_ringbuffer_get_size (ringbuffer);
 	int offset = totalsize - nbytes;
 
-	if ((nbytes / totalsize) > 0)
+	if ((totalsize != 0) && (nbytes / totalsize) > 0) {
```
kaixiong, line 429: Since totalsize is positive (when not zero), we can simplify the condition
```suggestion
	if (totalsize != 0 && nbytes > 0) {
```

hartwork: @kaixiong could it be you meant `nbytes >= totalsize` rather than `nbytes > 0`? What is the idea behind `nbytes > 0`?

hartwork:
> @hartwork Nice catch and fix. It looks like we can go one step further and eliminate the outer while loop altogether.

@kaixiong I see a risk of breaking things while doing that, I cannot say I understand all of that loop, I only touched the part that became clear and broken. If it's important, please elaborate in more detail.

libvisual/libvisual/lv_ringbuffer.c
kaixiong: @hartwork, I'm applying a bit of mathematics here. `nbytes / totalsize` is positive if and only if `nbytes` is positive.

Let $t$ be `totalsize`, $n$ be `nbytes`.

For $\frac{n}{t} > 0$ to hold, $n$ and $t$ must have the same sign. Since $t > 0$ , it follows that $n > 0$.

Alternatively, you can manipulate the inequality like this:
$\frac{n}{t} > 0 \implies n > 0 \times t \implies n > 0$
(Multiply both sides by $t$, the > condition is kept because $t > 0$)

hartwork: @kaixiong I see implications here rather than _equality_. `(nbytes / totalsize) > 0` in intergers is equal to `(nbytes / totalsize) >= 1` is equal to `nbytes >= totalsize`. `nbytes > 0` requires less from `nbytes`, it's not equivalent and I'm unsure why that's a beneficial change here. What do you think?

kaixiong:
> @kaixiong I see a risk of breaking things while doing that, I cannot say I understand all of that loop, I only touched the part that became clear and broken. If it's important, please elaborate in more detail.

@hartwork, that's prudent. Here's my reasoning:

- The inner loop can exit in two ways, via a `return`, via a `break`, or having failed the iteration condition `entry != NULL`.
- For `return` exits, the outer `while` loop never runs again.
- Since the loop has no `break` statements, non-return exits must have failed the `entry != NULL` condition, ergo `entry == NULL` is true.
- If `entry == NULL` then, the newly added `if` block always executes and returns, so the outer `while` loop is never run again.
- Conclusion: The outer `while` loop is not run a second time in all cases i.e. it is redundant.

libvisual/libvisual/lv_ringbuffer.c
kaixiong: @hartwork, I'm only working from the given premise of `(nbytes / totalsize) > 0`.

Your condition is correct too, but more complicated. The reason it's also correct is that we have _integer divisions_.

hartwork: @kaixiong I don't understand. Let's look at an example: for `nbytes = 3, totalsize = 5` we get different results after the suggested patch:

- Before: `(nbytes / totalsize) > 0` yields `false` (because of integer divison)
- After: `nbytes > 0` yields `true`

What am I missing?

kaixiong: @hartwork oh wait, my version is wrong because of integer division. You're right, my bad :sweat:

hartwork: @kaixiong no problem. Leave that part as-is then or `nbytes >= totalsize`? (I have a feeling that case `offset == totalsize` may need a closer look as well, but maybe not right now.)

kaixiong: @hartwork, yeah let's use `nbytes >= totalsize`. I also agree that it needs a closer look. In fact, I think the whole module could use a closer look.

Anyway, I got a bit too excited to have found an excuse to use LaTeX on Github..

hartwork: @kaixiong I noticed now that `nbytes > totalsize` (rather than `nbytes >= totalsize`) will save us for free in the `totalsize == nbytes` case because `offset := 0` is healthy but `offset := totalsize` may be not…

hartwork:
> @hartwork, that's prudent. Here's my reasoning:
> 
>  * The inner loop can exit in three ways, via a `return`, via a `break`, or having failed the iteration condition `entry != NULL`.
> 
>  * For `return` exits, the outer `while` loop never runs again.
> 
>  * Since the loop has no `break` statements, non-return exits must have failed the `entry != NULL` condition, ergo `entry == NULL` is true.
> 
>  * If `entry == NULL` then, the newly added `if` block always executes and returns, so the outer `while` loop is never run again.
> 
>  * Conclusion: The outer `while` loop is not run a second time in all cases i.e. it is redundant.

@kaixiong that's a great explanation, convinced and pushed. Also using `visual_mem_set` now, and a minimal-diff approach.

kaixiong (approved):

kaixiong:
> @kaixiong that's a great explanation, convinced and pushed. Also using `visual_mem_set` now, and a minimal-diff approach.

@hartwork Thanks. Good idea to use `visual_mem_set()` too.

PR #188, opened 2023-01-19
[0.4.x] libvisual-plugins: configure.ac: Drop unused X11 and make OpenGL logic respect pkg-config (e.g. for macOS)

hartwork:


kaixiong (commented):
@hartwork The patch looks good. I have some questions though.

-  I only just found out from this patch that the Mesa dev packages on my system distribute `gl.pc` and `glu.pc`. Is it also being use on macOS? 
- There are the macros `AX_CHECK_GL` and `AX_CHECK_GLU` from Autoconf archives. Would these be better?
https://www.gnu.org/software/autoconf-archive/ax_check_gl.html
https://www.gnu.org/software/autoconf-archive/ax_check_glu.html

kaixiong (approved):
@hartwork. All good, let's defer the use of `AX_CHECK_GL` and `AX_CHECK_GLU` to another PR.

PR #192, opened 2023-01-20
Core (Docs): Update links and C version requirement in readme files

kaixiong:
Update readme files. After a decade, ORC has been brought under the GStreamer project (?), and every website has HTTPS.

@hartwork, I am wondering if we should remove README and keep only README.md.


hartwork (approved):
> @hartwork, I am wondering if we should remove README and keep only README.md.

@kaixiong merging or deleting, would need to have a closer look, but yes :+1:

PR #197, opened 2023-01-21
[0.4.x] Add .gitignore files

hartwork:


kaixiong (commented):
@hartwork, would it be better to have one `.gitignore` at the top level?

hartwork:
> @hartwork, would it be better to have one `.gitignore` at the top level?

@kaixiong if the goal is to keep these clean over time, then maybe a single file doesn't scale and age as well. Would you strongly prefer one flat file for all?

kaixiong (approved):
@hartwork, I was thinking that we could have a top level ignore file with the commonalities and introduce more specific ones for core and plugins if needed. But I guess this makes the applied ignore list a bit harder to track down. I'm okay with this if you want to proceed.

hartwork:
@kaixiong there is two pairs of files that currently have (some) duplication:
- `libvisual{,-plugins}/.gitignore`
- `libvisual{,-plugins}/po/.gitignore`

We could resolve that duplication by pulling their shared entries up to the root `.gitignore` without specyging a path and hence ignore these files everywhere. It would be these entries:

From  `libvisual{,-plugins}/.gitignore`:
```
ABOUT-NLS
aclocal.m4
compile
config.guess
config.h
config.h.in
config.h.in~
config.log
config.rpath
config.status
config.sub
configure
configure~
depcomp
INSTALL
install-sh
libtool
ltmain.sh
missing
stamp-h1
```

And from `libvisual{,-plugins}/po/.gitignore`:
```
en@boldquot.header
en@quot.header
insert-header.sin
Makefile.in.in
Makevars
Makevars.template
POTFILES
remove-potcdate.sin
Rules-quot

*.gmo
```

Would that make a difference in a positive way, in your eyes?

kaixiong:
@hartwork Thanks, it does look like an improvement.

What causes the files `configure.h.in~` and `configure~` to be generated? I'm thinking that we could just use `*~`.

hartwork:
> What causes the files `configure.h.in~` and `configure~` to be generated?

@kaixiong not sure, I just grepped for it and could find it and it wasn't me manually and non of my tools that I would be aware :shrug: 

> I'm thinking that we could just use `*~`.

Maybe that's a good idea, done :+1: 

> @hartwork Thanks, it does look like an improvement.

Okay nice, I have applied that refactoring now, the diff can probably be best seen at https://github.com/Libvisual/libvisual/compare/84cac7dbabf3239eb773c33d381bc6caa72d0a22..1e03d4f793764f787f927e65c27a25a4e3615bed .

kaixiong (approved):

PR #199, opened 2023-01-21
[0.4.x] libvisual-plugins: Fix out-of-bounds access for actor "lv_flower"

hartwork:


libvisual-plugins/plugins/actor/pseudotoad_flower/actor_flower.c
```diff
@@ -280,14 +280,14 @@ int lv_flower_render (VisPluginData *plugin, VisVideo *video, VisAudio *audio)
 		float y;
 		int i;
 		for (i=0; i<priv->nof_bands; i++) {
-			y=temp_bars[i * 8];
+			y=temp_bars[(i * 8) % NOTCH_BANDS];
```
kaixiong, line 283: I see that this fixes the overflow. But could this line have been intended to be:
```c
y=temp_bars[i];
```

kaixiong (commented):
Nice catch! I have a question though.

libvisual-plugins/plugins/actor/pseudotoad_flower/actor_flower.c
kaixiong: Or perhaps:

```cpp
y=temp_bars[i] * 8;
```

hartwork: @kaixiong good question! I tried `temp_bars[i] * 8;` and then the flower goes even more crazy while its already jumping around a lot so I ruled out that as intended. `temp_bars[i];` was the other candidate.

I checked Git history now and found that:
- `float temp_bars[NOTCH_BANDS];` was first introduced in commit 939266aa5b31d540c367408e812f8d1e2a9b9248.
- The commit replaces `y=freqnorm[i * 8];` by `y=temp_bars[i * 8];`.
- However `freqnorm` is defined as `float freqnorm[256];` was used, which is 8 times bigger.
- Hence `i * 8` likely should have been `i` in that commit, the first place.

What do you think?

kaixiong: @hartwork, yeah I guess we will fix it to just `i` instead.

I kind of threw in the last bit about `temp_bars[i] * 8`. The flower already looks correctly scaled, magnifying it by 8 times is nuts to say the least.

hartwork: @kaixiong done

kaixiong (approved):

PR #201, opened 2023-01-22
[0.4.x] lv-tool: Always skip actors "gstreamer" and "gdkpixbuf" when cycling

hartwork:
Inspired by https://github.com/hartwork/xmms2-devel/blob/2b8cc16791aceaf6ff121418041f1f6ae9f99c86/src/clients/vistest/libvisual.c#L142 .

### DIFF
--- a/libvisual/po/de.po
+++ b/libvisual/po/de.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: libvisual 0.4.0-2.2\n"
 "Report-Msgid-Bugs-To: https://github.com/Libvisual/libvisual/issues/\n"
-"POT-Creation-Date: 2023-01-19 17:24+0100\n"
+"POT-Creation-Date: 2023-01-22 04:12+0100\n"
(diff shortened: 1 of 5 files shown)


kaixiong (commented):
@hartwork Good idea! I'm thinking that we should generalise this slightly by using an array of actors to skip.

hartwork:
> @hartwork Good idea! I'm thinking that we should generalise this slightly by using an array of actors to skip.

@kaixiong no objections in general, but before I have brushed up on C++ more this will take me more time than it adds practical value, at least today, and I can bring more value to other pull requests with plain C using that same time. Would you mind taking over this C++ refactoring?

kaixiong:
> @kaixiong no objections in general, but before I have brushed up on C++ more this will take me more time than it adds practical value, at least today, and I can bring more value to other pull requests with plain C using that same time. Would you mind taking over this C++ refactoring?

@hartwork. Yup, I can do it. Let's proceed with this then. The PR queue is getting long :)

kaixiong (approved):

hartwork:
@kaixiong thank you! :+1:

PR #200, opened 2023-01-21
[0.4.x] libvisual-plugins: Fix out-of-bounds access for actor "infinite"

hartwork:
Note that function `_inf_load_effects` is currently called multiple times.

PR #202, opened 2023-01-23
libvisual-plugins: Fix out-of-bounds access for actor "infinite"

hartwork:
Foward-port of 0.4.x pull request #200


libvisual-plugins/plugins/actor/infinite/display.c
```diff
@@ -162,15 +162,28 @@ static void _inf_compute_surface(InfinitePrivate *priv, t_interpol* vector_field
...
 				interpol = &vector_field[add_dest];
-				add_src = (interpol->coord & 0xffff) * priv->plugwidth + (interpol->coord >> 16);
-				ptr_pix = priv->surface1 + add_src;;
+				const uint16_t y = interpol->coord & 0xffff;
+				const uint16_t x = interpol->coord >> 16;
+				(void)add_src;
+				ptr_pix = priv->surface1 + y * priv->plugwidth + x;
 
-				/* FIXME it does buffer overread here now and then */
+				const uint8_t *ptr_pix_end = priv->surface1 + (priv->plugwidth * priv->plugheight);
```
kaixiong, line 171: @hartwork Not sure if I missed this for master, but this line should be hoisted out of the rendering loop. `ptr_pix_end` depends only on the bitmap data pointer and dimensions, which does not change in the loop.

The compiler should be able to easily optimize loop invariants like these, but I think it makes sense organizationally to have the computation done before the loop. It also makes debug builds faster.

kaixiong (changes requested):
Just one minor change to make.

libvisual-plugins/plugins/actor/infinite/display.c
hartwork: @kaixiong alright, done and pushed. I'll backport to 0.4.x after the merg.

PR #206, opened 2023-01-23
Core (LV::Video): Implement saves to PNG files

kaixiong:


libvisual/libvisual/lv_video.cpp
```diff
@@ -259,6 +263,21 @@ namespace LV {
       return m_impl->buffer->is_allocated ();
   }
 
+  bool Video::save_to_file (std::string const& path) const
+  {
+      auto extension = fs::path {path}.extension ();
+
+      std::ofstream file {path};
```
hartwork, line 270: I believe this^^ line needs to be moved into the `if (extension == ".png")` block 0r else an empty file will have been created already.

kaixiong: @hartwork, nice catch. I was already tackling this albeit indirectly, thinking of the scenario where the file creation failed instead.

kaixiong: @hartwork This code was only just ported from a graphics engine I'm writing on the side. The port is still untested.

At the moment, I'm trying to build up the testing support a little so I can write unit tests for this and other `LV::Video` functions. In the works are two methods for `LV::Video`, `validate()` and `has_same_content()` for validating the internal consistency of an instance, and for comparing two instances respectively.

hartwork: > @hartwork This code was only just ported from a graphics engine I'm writing on the side. The port is still untested.

@kaixiong how do you feel about using either [draft status](https://github.blog/2019-02-14-introducing-draft-pull-requests/) or a subject prefix like "[W.I.P.] " for cases like this? I think that that would make it quicker to see (without need to remember) for cases like this.

kaixiong: @hartwork Done. This was the intention but I forgot to set the draft status it seems.

kaixiong:
@hartwork Looks like I somehow confused the branches, lost the changes locally,  then lost it from GitHub via a forced push. Is it still possible to recover it from GitHub?

hartwork:
@kaixiong I'm optimistic that that can be fixed. A bit further up GitHub shows this:

Screenshot_20230205_180543_git_compare

Following the [link behind that compare button](https://github.com/Libvisual/libvisual/compare/c2b91ebffece693826fb3868ba1834a892285e6f..ed61ea41f452f0de547c12e661e4580a71a7a899) shows that the old commit still had PNG saving code in it. I think what you'll need for a fix is something like this:
```bash
git checkout png-saves
git reset --hard c2b91ebffece693826fb3868ba1834a892285e6f
git push origin png-saves --force-with-lease
```
If we did not have that help from GitHub, I'd try something like `git reflog --all | grep 'saving to PNG'` locally. It doesn't show output here, but since you have more local history on that branch than me, maybe it shows things for you?

What do you think?

kaixiong:
@hartwork Thanks, that seems to have worked. I hadn't realised there was a compare button in the GitHub timeline. Wonder what actually went down.

hartwork:
@kaixiong good to see PNG saving back! :+1:

libvisual/libvisual/private/lv_video_png.cpp
```diff
@@ -153,27 +191,126 @@ namespace LV {
...
+          case VISUAL_VIDEO_DEPTH_24BIT: {
+              bit_depth = 8;
+              color_type = PNG_COLOR_TYPE_RGB;
+              break;
+          }
+          case VISUAL_VIDEO_DEPTH_32BIT: {
+              bit_depth = 8;
+              color_type = PNG_COLOR_TYPE_RGB_ALPHA;
+              break;
+          }
```
hartwork, line 271: `bit_depth` was unexpected to me here. It seems to be per channel. My vote for renaming to `bits_per_channel` or similar — something that makes it clear that it's not per image but per color channel.

libvisual/libvisual/private/lv_video_png.cpp
```diff
@@ -153,27 +191,126 @@ namespace LV {
...
+      auto row_stride {png_get_rowbytes (png_ptr, info_ptr)};
 
       // NOTE: We have to use visual_mem_malloc() here as LV::Video
       // will free the buffer with visual_mem_free()
       pixels = static_cast<uint8_t*> (visual_mem_malloc (row_stride * height));
 
       pixel_row_ptrs = new uint8_t*[height];
 
-      for (unsigned int y = 0; y < height; y++) {
+      for (unsigned int y {0}; y < height; y++) {
```
hartwork, line 204: (I'm biased here being a lot closer to C than C++ myself but `unsigned int y {0}` is taking C++ to far in my view: it looks super weird and hurts readability at least with me, and there are other places in this pull request that do use `int y = 0` in `for` loops where this new pattern has not been applied. (I haven't checked carefully if that code was moved or added.) It's not super important to me but for plain integers I consider this a step backwards, just my 2 cents.)

hartwork (changes requested):
@kaixiong found some time now, here's some minor things I found:

libvisual/libvisual/private/lv_video_png.cpp
kaixiong: I agree, it's not my favoured syntax and takes a while to get used to (even as a fairly seasoned C++ guy) but I've started using uniform initialization even for numeric types. The reason is that unlike the alternatives, `{}` will never perform implicit conversions except widening.

It's not important in this particular instance, unlike numerical code. I've used it here mostly for consistency.

To elaborate a bit:

```C++
int x {y};   // y must be int or narrower

int x = y;   // narrowing can occur if y is long, float, etc. static_cast<> not necessary
             // other implicit conversions possible e.g. via conversion constructors 

int x (y);   // narrowing can occur if y is long, float, etc. static_cast<> not necessary
             // other implicit conversions possible e.g. via conversion constructors
```

This is also true when declaring or constructing values of other types, so it's a good way to add some clarity (at the cost of initial unreadability?). The rest of this source file adopts this as well.

libvisual/libvisual/private/lv_video_png.cpp
kaixiong: Yes, good idea.

libvisual/libvisual/lv_video.cpp
```diff
@@ -36,19 +36,55 @@
...
+
   namespace {
 
+    using BitmapLoad = std::function<VideoPtr (std::istream&)>;
+    using BitmapSave = std::function<bool (Video const&, std::ostream&)>;
+
+    std::unordered_map<std::string, BitmapLoad> const bitmap_load_map =
+    {
+        { "bmp", bitmap_load_bmp },
+        { "png", bitmap_load_png }
```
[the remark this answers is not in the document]
kaixiong: Hmm, I don't really see the value of enums.

These are just table keys (and fairly descriptive at that) that only the loader needs to care about. They're meant to link the tables `bitmap_save_map`, `bitmap_load_map` and `bitmap_format_extension_map`. 

Named constants could help prevent typos but then mistakes will be caught by the unit tests.

libvisual/libvisual/private/lv_video_png.cpp
hartwork: @kaixiong thanks for elaborating :+1:

libvisual/libvisual/lv_video.cpp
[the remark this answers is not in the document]
hartwork: > Hmm, I don't really see the value of enums.

There are two aspects:
- typo detection now happens at compile time rather than (app or test) runtime
- all places using the magic value would now communicate that it's about a file format rather than merely a file extension (thinking of e.g. `FILE_FORMAT_PNG` in naming). Given that we're dealing with extensions some places and formats in others, that separation would also help clarity in my view.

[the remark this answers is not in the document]
kaixiong: Ah you mean using enum values for keys instead. Yes, much nicer. Guards against typos and more efficient lookups too.

I've forgotten about the use of the format ID in the added `save_to_stream()` method, it definitely is visible to API users.

Let me make that change.

hartwork (approved):

PR #209, opened 2023-01-23
[0.4.x] libvisual-plugins: Fix out-of-bounds access for actor "oinksie"

hartwork:


kaixiong (commented):
@hartwork Additions modulo n is a very common operation, so I suggest having a function called `mod_add()` instead, implemented like this:

```c
int mod_add(int a, int b, unsigned int n)
{
    int c = (a + b) % n;
    return c >= 0 ? c : c + n;
}
```

hartwork:
@kaixiong if we do that we'll have to avoid overflow in there, because the function is generic, while the values in oinksie seem to be within (-10,000, +10,000). Also we'd need to to put it elsewhere to be reused. Is it really with it?

kaixiong:
@hartwork, my idea is that we should express the calculation using a rather common operation. The intention is immediately clear:

```c
xi = (_oink_table_sin[mod_add(xturn, add1, OINK_TABLE_NORMAL_SIZE)] * (float) (priv->screen_width / (number + 1)));
yi = (_oink_table_cos[mod_add(yturn, add2, OINK_TABLE_NORMAL_SIZE)] * (float) (priv->screen_height / 5));
```

There is a surprising mistake in `mod_add()` - `n` needs to be an `int` (not sure why):
```c
int mod_add(int a, int b, int n)
{
    int c = (a + b) % n;
    return c >= 0 ? c : c + n;
}
```

This way of adjusting the remainder is also more intuitive in my opinion.

I just noticed that `ring_mod()` as it is implemented in the PR yields wrong results. For example, `ring_mod(-5, 5)` gives 4. It should be 0. Below is a quick test I wrote up. `mod(a, n)` is essentially implemented as `mod_add(a, 0, n)`.

```
-6 % 5 = -1
ring_mod (-6,5) = 5
     mod (-6,5) = 4
-5 % 5 = 0
ring_mod (-5,5) = 4
     mod (-5,5) = 0
-4 % 5 = -4
ring_mod (-4,5) = 3
     mod (-4,5) = 1
-3 % 5 = -3
ring_mod (-3,5) = 2
     mod (-3,5) = 2
-2 % 5 = -2
ring_mod (-2,5) = 1
     mod (-2,5) = 3
-1 % 5 = -1
ring_mod (-1,5) = 5
     mod (-1,5) = 4
0 % 5 = 0
ring_mod (0,5) = 0
     mod (0,5) = 0
1 % 5 = 1
ring_mod (1,5) = 1
     mod (1,5) = 1
2 % 5 = 2
ring_mod (2,5) = 2
     mod (2,5) = 2
3 % 5 = 3
ring_mod (3,5) = 3
     mod (3,5) = 3
4 % 5 = 4
ring_mod (4,5) = 4
     mod (4,5) = 4
5 % 5 = 0
ring_mod (5,5) = 0
     mod (5,5) = 0
6 % 5 = 1
ring_mod (6,5) = 1
     mod (6,5) = 1
```

hartwork:
@kaixiong peplying to your latest comments first:

> @hartwork Regarding your concern about overflow in the general case, it would have to be worked around with something like `mod(mod(a,n) + mod(b,n), n)` one way or another. Or slightly simpler if `a` is already in 0..n-1, with `mod(a + mod(b,n), n)`.
> 
> Not sure anyone would like to write any of that if they could help it. So, I think it's reasonable to require that `a + b` not overflow if `mod_add(a,b,n)` were a library function for general use.

The more this becomes a library function, the more it has to handle overflow inside, I strongly disagree.

> But this is a discussion for another time since the numbers we are dealing with here are relatively small.

If reusability is out the door, we don't need to pull addition in.

> @hartwork, my idea is that we should express the calculation using a rather common operation. The intention is immediately clear:
> 
> ```c
> xi = (_oink_table_sin[mod_add(xturn, add1, OINK_TABLE_NORMAL_SIZE)] * (float) (priv->screen_width / (number + 1)));
> yi = (_oink_table_cos[mod_add(yturn, add2, OINK_TABLE_NORMAL_SIZE)] * (float) (priv->screen_height / 5));
> ```

To be honest I find `ring_mod(xturn + add1, OINK_TABLE_NORMAL_SIZE)` more clear.
I'm not totally against `mod_add` from a usage point of view, but once we add overflow protection it will become quite complex. Is this about true reuse? Where would we put a truly reusable implementation and make sure it's still inlineable if the compile likes to do that? Would it need to be a macro? There's a pile of questions here where I find it's at least questionable if that is within scope, unless we have true reuse.
Do we agree that `mod_add` needs a good place and overflow protection and inlinability to be of real use? Else let's stick to a fixed version of `ring_mod` please.

> There is a surprising mistake in `mod_add()` - `n` needs to be an `int` (not sure why):

Yes, my comment has that covered:
```c
// In C, `-8 % 5` yields -3 (and `-8 % 5u` yields +3) rather than 2. This wrapper corrects for it.
```
In short, the signedness of the right operand matters.

> 
> ```c
> int mod_add(int a, int b, int n)
> {
>     int c = (a + b) % n;
>     return c >= 0 ? c : c + n;
> }
> ```
> 
> This way of adjusting the remainder is also more intuitive in my opinion.

I don't really agree but I have not problem with making it closer to that version. Updated.

> I just noticed that `ring_mod()` as it is implemented in the PR yields wrong results. For example, `ring_mod(-5, 5)` gives 4. It should be 0. Below is a quick test I wrote up. `mod(a, n)` is essentially implemented as `mod_add(a, 0, n)`.

Good catch. `(a < 0)` would need to be `(a < 0) && (original != 0)` in my version. Fixed.

hartwork:
@kaixiong PS: I see that `_oink_table_sin[abs (tab % OINK_TABLE_NORMAL_SIZE)]` is used elsewhere, so mirroring was considered acceptable or preferable in the past (instead of fully sticking to a ring). We can also resort to using `abs` here, if you prefer. What do you think?

kaixiong:
@hartwork, I'm fine with having `ring_mod()` but the name should really be changed to `mod()`. Ring means something (an algebraic structure) in mathematics and while it can apply here, it's overly specific. Technically, we're working with the additive integer group $Z_n$.

My method of adjusting the remainder is based on the following idea:

C's `%` operator gives the correct result for negative numbers modulo n. For n = 5, we will always get a number from -4 to 0. Adding 5 brings it back into the 0..4 range. Adding or subtracting 5 (or any multiple of 5) in modulo 5 is the equivalent of adding a zero, so you only need to take care to make sure the final result is in the right range i.e. 0..4.

hartwork:
> @hartwork, I'm fine with having `ring_mod()` but the name should really be changed to `mod()`. Ring means something (an algebraic structure) in mathematics and while it can apply here, it's overly specific. Technically, we're working with the additive integer group Zn.

Thank you! Name adjusted, pushed. `ring_` was in fact mean to reference that very thing from algebra. It's called "Restklassenring" in German. I checked https://en.wikipedia.org/wiki/Ring_(mathematics) now and it seems to be the same.

> My method of adjusting the remainder is based on the following idea:
> 
> C's `%` operator gives the correct result for negative numbers modulo n. For n = 5, we will always get a number from -4 to 0. Adding 5 brings it back into the 0..4 range. Adding or subtracting 5 (or any multiple of 5) in modulo 5 is the equivalent of adding a zero, so you only need to take care to make sure the final result is in the right range i.e. 0..4.

I think you're saying that your motivation was puristic and mathematical. If that's not what you meant, please help me understand better.

hartwork:
@kaixiong interesting point :+1:

kaixiong:
> Thank you! Name adjusted, pushed. `ring_` was in fact meant to reference that very thing from algebra. It's called "Restklassenring" in German. I checked https://en.wikipedia.org/wiki/Ring_(mathematics) now and it seems to be the same.

@hartwork Cool! The reason why I would prefer not to bring rings into the picture is that a ring also has a multiplicative operator. At least based on the snippets I'm seeing, we are only performing additions.

> I think you're saying that your motivation was puristic and mathematical. If that's not what you meant, please help me understand better.

It's nothing puristic (I cringe at that word!). My approach is based on elementary properties of modular arithmetic. 

Think of the timeline. Then think of the minute hand's position on the analog clock. Adding or subtracting any number of hours to a given time does not change the minute position. For example, the 1st minute, 61st minute and 121st minute  considered to be the same. This is essentially arithmetic with modulus 60. If you work with degrees, you would have a modulus of 360. Generalizing this to _n_ we have this property of addition:
```
mod(a + b, n) == mod(a,n) + mod(b,n)
```
Note that the equality is defined similarly to the equivalence of minute hand positions on the clock.

We also have:
```
mod(n, n) = 0
```
So putting them together, we have:

```
mod(a + kn, n)                 (k is an integer)
== mod(a, n) + mod(kn, n)
== mod(a, n) + [mod(n, n) + mod(n, n) + ...]  (up to k times)
== mod(a, n) + [0 + 0 + ...]   (up to k times)
== mod(a, n)
```

So if you receive a correct value at the end of a computation  in the wrong range, like an angle of 370 degrees or -370 degrees, you can always add or subtract 360 degrees until the final result falls between 0 and 360 degrees.

Going back to the original problem, we know that `x % n` yields a number in the range `-(n-1) .. -1` when `x < 0`. To get the result back in the range `0 .. n-1`, we just need to add n once.

kaixiong (approved):

hartwork:
@kaixiong should do we do any more adjustments to the current patch to be good enough to merge?

kaixiong:
@hartwork, nah it's all good :+1:

I did not expect this extended discussion :laughing:. I respect your point about an `mod_add()` implementation ensuring no overflows. It's a design philosophy focused on correctness i.e. functions should be total (as opposed to partial) and must work for all inputs.

I'm advocating from a real-time graphics programming point of view, which is performance focused. If functions needed to be this airtight, especially when working with floating point values, I'd have to do numerical analysis each time!

PR #215, opened 2023-01-24
lv-tool: Fix check for a suitable palette

hartwork:
We can only copy 256 entries if we have them...

(based on 0.4.x commit d1edda999f2609537d983e6ed65b843d4a02a8d5 from pull request #213)

### DIFF
--- a/libvisual/tools/lv-tool/display/sdl_driver.cpp
+++ b/libvisual/tools/lv-tool/display/sdl_driver.cpp
@@ -226,7 +226,7 @@ namespace {
           if (m_screen->format->BitsPerPixel == 8) {
               auto const& pal = m_display.get_video ()->get_palette ();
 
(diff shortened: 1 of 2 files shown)


kaixiong (commented):
@hartwork, I don't think this patch is correct. It's clearly intended to copy palettes with smaller than 256 entries and fill the remaining with black.

What's slightly questionable in this implementation is the use of visual_mem_set() to zero out the entries but it should be safe since SDL_Color is just a regular C struct defined with uint8_t (unless I'm missing something about std::array's guarantees regarding memory layout).

hartwork:
> @hartwork, I don't think this patch is correct. It's clearly intended to copy palettes with smaller than 256 entries and fill the remaining with black.

@kaixiong you're right, sorry for not having checked in depth. Thanks for catching this! :+1: Closing…

PR #219, opened 2023-01-25
[0.4.x] lv-tool: Resolve strcmp duplication around actors to always exclude from cycling

hartwork:
Follow-up to pull request #201 to prepare/precede porting to `master`

@kaixiong some choices that were intentional (but are up for discussion) are:
- I picked `set` over `vector` because it seemed to capture intent better and performance is of no concern with only 2 entries.
- I picked `set` over `unordered_set` to not require C++11.
- I picked `set.find` over `set.contains` to not require C++20.
- I chose uppercase because the data is constant in nature.
- And put it outside the function so the code is not run more times than needed.

PS: What's a good requirement for a C++ standard for 0.4.x when ideally [everyone who is on <0.4.2 today](https://repology.org/project/libvisual/badges) should be able to update? C++03, 11, 14, 17?  C++17 [seems to require](https://en.wikipedia.org/wiki/C%2B%2B17#Compiler_support) Clang 5, GCC 8 or MSVC 2017, with GCC 8.1 released [in early 2018](https://gcc.gnu.org/gcc-8/), the latest of these three I believe. So not strictly 5 years ago, if that was the goal.

What do you think?

### DIFF
--- a/libvisual/po/de.po
+++ b/libvisual/po/de.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: libvisual 0.4.0-2.2\n"
 "Report-Msgid-Bugs-To: https://github.com/Libvisual/libvisual/issues/\n"
-"POT-Creation-Date: 2023-01-22 13:50+0100\n"
+"POT-Creation-Date: 2023-01-27 01:58+0100\n"
 "PO-Revision-Date: 2017-08-10 18:46+GMT\n"
 "Last-Translator: Chris Leick <c.leick@vollbio.de>\n"
 "Language-Team: German <debian-l10n-german@lists.debian.org>\n"
@@ -900,7 +900,7 @@ msgstr ""
 msgid "Invalid depth passed to the scaler"
 msgstr "Ungültige Tiefe an Scaler übergeben"
 
-#: tools/lv-tool/lv-tool.cpp:813 tools/lv-tool/lv-tool.cpp:910
+#: tools/lv-tool/lv-tool.cpp:819 tools/lv-tool/lv-tool.cpp:916
 msgid "lv-tool"
 msgstr ""
 
--- a/libvisual/po/es_AR.po
+++ b/libvisual/po/es_AR.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Libvisual Library 0.4.0\n"
 "Report-Msgid-Bugs-To: https://github.com/Libvisual/libvisual/issues/\n"
-"POT-Creation-Date: 2023-01-22 13:50+0100\n"
+"POT-Creation-Date: 2023-01-27 01:58+0100\n"
 "PO-Revision-Date: 2005-03-20 17:09-0300\n"
 "Last-Translator:  <dprotti@users.sourceforge.net>\n"
 "Language-Team: Spanish\n"
@@ -918,7 +918,7 @@ msgstr ""
 msgid "Invalid depth passed to the scaler"
 msgstr "Profundidad pasada al escalador inválida"
 
-#: tools/lv-tool/lv-tool.cpp:813 tools/lv-tool/lv-tool.cpp:910
+#: tools/lv-tool/lv-tool.cpp:819 tools/lv-tool/lv-tool.cpp:916
 msgid "lv-tool"
 msgstr ""
 
--- a/libvisual/po/es_ES.po
+++ b/libvisual/po/es_ES.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Libvisual Library 0.4.0\n"
 "Report-Msgid-Bugs-To: https://github.com/Libvisual/libvisual/issues/\n"
-"POT-Creation-Date: 2023-01-22 13:50+0100\n"
+"POT-Creation-Date: 2023-01-27 01:58+0100\n"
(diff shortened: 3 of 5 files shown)


kaixiong:
@hartwork

>     * I picked `set` over `vector` because it seemed to capture intent better and performance is of no concern with only 2 entries.

I thought anything but `vector`/`array` was way overkill but `set` does make sense. Not sure how much 'bloat' bringing in set introduces (it's a template after all). Internally it must have a far more complex implementation than a `vector` or `array`. For example, `libstdc++` uses a red-black tree. I personally prefer `array`.

>     * I picked `set.find` over `set.contains` to not require C++20.

Makes sense. I was a bit surprised that `contains()` didn't arrive until C++20. master already requires C++20, so I would prefer the version on master to use `contains()` if we're sticking to `set`.

Do note the functions are called `set::find()` and `set::contains()` respectively, rather than `set.find()` and `set.contains()`.

Also, when initializing a container with an initializer list you should drop the = like this:

```c++
inline const std::set<std::string> actors_to_exclude_from_cycling { ... };
```
_`inline const` is the equivalent of `static const` in C++_

>     * I chose uppercase because the data is constant in nature.

Very big no from me. I don't think I've seen a C++ codebase using uppercase for named constants. All uppercase names should be reserved for preprocessors.

If you like to differentiate them somehow - I don't see a need to - a nice technique I've seen suggested is adding a namespace, like `Constants`.

```c++
namespace Constants
{
  inline const std::set<std::string> actors_to_exclude_from_cycling { ... };
  // ...
} 

// ...

if (Constants::actors_to_exclude_from_cycling.contains(actor_name)) {
  // ...
}
```

Apart from that, I find the name a bit too long.

>     * And put it outside the function so the code is not run more times than needed.

Yes, it should be. 

> 
> PS: What's a good requirement for a C++ standard for 0.4.x when ideally [everyone who is on <0.4.2 today](https://repology.org/project/libvisual/badges) should be able to update? C++03, 11, 14, 17? C++17 [seems to require](https://en.wikipedia.org/wiki/C%2B%2B17#Compiler_support) Clang 5, GCC 8 or MSVC 2017, with GCC 8.1 released [in early 2018](https://gcc.gnu.org/gcc-8/), the latest of these three I believe. So not strictly 5 years ago, if that was the goal.
> 
> What do you think?

I think C++11 is good enough for 0.4.x. The main concern I see is backporting updates from master (which is on C++20). Even then I believe the adaptations needed should be minimal, compared to say C++11 against C++98.

hartwork:
> Makes sense. I was a bit surprised that `contains()` didn't arrive until C++20. master already requires C++20, so I would prefer the version on master to use `contains()` if we're sticking to `set`.

@kaixiong ack.


> Do note the functions are called `set::find()` and `set::contains()` respectively, rather than `set.find()` and `set.contains()`.

Alright!

> Also, when initializing a container with an initializer list you should drop the `=` like this:
> 
> ```c++
> inline const std::set<std::string> actors_to_exclude_from_cycling { ... };
> ```

I put it in by accident and dropped it now. Learned that it [requires C++11](https://i.stack.imgur.com/aoQbA.png).


> _`inline const` is the equivalent of `static const` in C++_

I don't really know either side of that in detail. Are you recommending to add inline here? Do you have good links on that? I found some C++17 things on the matter but I'm not sure what to make it of it.

> * I chose uppercase because the data is constant in nature.
> 
> Very big no from me. I don't think I've seen a C++ codebase using uppercase for named constants. All uppercase names should be reserved for preprocessors.

Okay :smiley: 

> If you like to differentiate them somehow - I don't see a need to - a nice technique I've seen suggested is adding a namespace, like `Constants`.
> 
> ```c++
> namespace Constants
> {
>   inline const std::set<std::string> actors_to_exclude_from_cycling { ... };
>   // ...
> } 
> 
> // ...
> 
> if (Constants::actors_to_exclude_from_cycling.contains(actor_name)) {
>   // ...
> }
> ```

Makes sense, but I don't see much value in `Constants::` here. I'm okay if all we do on constants is `const`.

> Apart from that, I find the name a bit too long.

Short and precise are often in conflict, in particular here where there is a another list of actors to exclude from cycling, except those can be controlled by the user while the ones here apply always, on top. I'm putting in short `actors_to_skip` now but it's not very precise.

> > PS: What's a good requirement for a C++ standard for 0.4.x when ideally [everyone who is on <0.4.2 today](https://repology.org/project/libvisual/badges) should be able to update? C++03, 11, 14, 17? C++17 [seems to require](https://en.wikipedia.org/wiki/C%2B%2B17#Compiler_support) Clang 5, GCC 8 or MSVC 2017, with GCC 8.1 released [in early 2018](https://gcc.gnu.org/gcc-8/), the latest of these three I believe. So not strictly 5 years ago, if that was the goal.
> > What do you think?
> 
> I think C++11 is good enough for 0.4.x. The main concern I see is backporting updates from master (which is on C++20). Even then I believe the adaptations needed should be minimal, compared to say C++11 against C++98.

Alright, C++11 it is! :+1:

kaixiong (approved):

kaixiong:
> > _`inline const` is the equivalent of `static const` in C++_
> 
> I don't really know either side of that in detail. Are you recommending to add inline here? Do you have good links on that? I found some C++17 things on the matter but I'm not sure what to make it of it.

It's just C++'s way of declaring that the symbol is file/internal linkage only. I don't think it actually matters here since it's not library code.

I just approved the PR :)

PR #220, opened 2023-01-26
C99 compatibility fixes to avoid implicit function declarations

fweimer-rh:
Strictly speaking, C99 no longer supports implicit function declarations.  Future compilers may treat them as errors (instead of issuing warnings), to avoid accidental parameter or return type mismatches.

Include "lv_cpu.h" for the visual_cpu_initialize, visual_cpu_get_sse, visual_cpu_get_3dnow functions.  Add a prototype for visual_transform_init, so that it can be called before it is defined.

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC

### DIFF
--- a/libvisual/libvisual/lv_libvisual.c
+++ b/libvisual/libvisual/lv_libvisual.c
@@ -30,6 +30,7 @@
 #include <gettext.h>
 
 #include "lvconfig.h"
+#include "lv_cpu.h"
 #include "lv_plugin.h"
 #include "lv_actor.h"
 #include "lv_input.h"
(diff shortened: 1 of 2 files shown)


hartwork (changes requested):
Hi @fweimer-rh,

thanks for taking this upstream!  When looking at the related changes in detail and the description I find that:
- We have Clang 16 by now in the CI which is one of those compilers that rejects lack of implicit prototypes.
- Given that we had to support the new strictness in the past, I checked again and found that we have already fixed thes very two issues in commit 178568d89d760f221c5b871f2c2f331ee645d4fa, i.e.
  - there already is an include to `"lv_cpu.h"` 10 lines down and
  - there already is a prototype for `visual_transform_init` 8 lines down.

I would close this pull request as "already fixed" unless I am missing something.

PR #221, opened 2023-01-27
lv-tool: Always skip actors "gstreamer" and "gdkpixbuf" when cycling

hartwork:
Forword-port of 0.4.x pull requests #201 and #219 but with `unordered_set::contains` of C++20.

### DIFF
--- a/libvisual/tools/lv-tool/lv-tool.cpp
+++ b/libvisual/tools/lv-tool/lv-tool.cpp
@@ -32,6 +32,7 @@
 #include <atomic>
 #include <stdexcept>
 #include <iostream>
+#include <unordered_set>
 #include <string>
 #include <cstdio>
 #include <cstdlib>
@@ -55,6 +56,11 @@
 #endif
 
 namespace {
+
+  const std::unordered_set<std::string> actors_to_skip {
+      "gdkpixbuf",
+      "gstreamer"
+  };
 
   std::string actor_name;
   std::string input_name = DEFAULT_INPUT;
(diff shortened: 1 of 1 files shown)


libvisual/tools/lv-tool/lv-tool.cpp
```diff
@@ -378,6 +379,11 @@ namespace {
       std::signal (SIGTERM, handle_termination_signal);
   }
 
+  const std::unordered_set<std::string> actors_to_skip {
```
kaixiong, line 382: @hartwork, I just noticed that this is stuck in between function definitions. It should be brought to the top of the file.

hartwork: @kaixiong the intention was to be in proximity, can sure be argued. I'd like to pull the "if #219 was good enough, then this" card here, e.g. because the idea was to keep the two copies close. Does it matter enough to you to be fixed on both sides?

kaixiong: @hartwork I would prefer it be fixed on both sides. Thanks in advance!

My strong preference is for globals, especially app settings, to be up top, rather than stuffed between function definitions. The centralisation makes them easier to find/view/tweak in one glance. This is also in keeping with the current organization of the file.

There are cases where proximity make more sense though, such as when the source file is sectioned into regions, one for each logical component. Other times, it would be extremely verbose or impossible to place a declaration at the beginning, like say for a lookup table, because that is initialized with a number of function calls, etc.

hartwork: > @hartwork I would prefer it be fixed on both sides. Thanks in advance!

@kaixiong alright, moved up :beers:

kaixiong (approved):

PR #226, opened 2023-01-29
Add LV::Video testing functions

kaixiong:
Added `LV::Video::validate()` and `LV::Video::has_same_content()` for unit testing video functions. I am building up the test infrastructure for various enhancements, fixes, and changes to `LV::Video`'s API that are on the pipeline. Here is a list:
- Saving to PNG
- Fixing of sub-videos
- Removal of API methods that mutate LV::Video atttributes


libvisual/libvisual/lv_video.cpp
```diff
@@ -292,6 +293,110 @@ namespace LV {
...
+      if (m_impl->parent) {
+          auto parent_impl = m_impl->parent->m_impl.get ();
+          if (!parent_impl->extents.contains (m_impl->extents)) {
+              visual_log (VISUAL_LOG_ERROR, "Sub-video is not fully contained by parent.");
+              return false;
+          }
+      }
+
+      if (m_impl->depth != VISUAL_VIDEO_DEPTH_GL) {
+          // Validate depth and bytes per pixel consisteency.
```
hartwork, line 341: Typo "consisteency" here^^

libvisual/libvisual/lv_video.cpp
```diff
@@ -292,6 +293,110 @@ namespace LV {
...
+
+          auto pixel_row_ptr = static_cast<uint8_t const*> (get_pixels ());
+
+          for (int y = 0; y < m_impl->height; y++) {
+              if (m_impl->pixel_rows[y] != pixel_row_ptr) {
+                  visual_log (VISUAL_LOG_ERROR, "Pixel row pointer table is wrong at y=%d.", y);
+                  return false;
+              }
+
+              pixel_row_ptr += m_impl->pitch;
```
hartwork, line 359: I'd inline this to `expected_pixel_row_ptr = pixel_row_ptr + y * m_impl->pitch` and compare against that. The current code only makes full sense once the reader arrives at line 359 which is arguably later than ideal.

libvisual/libvisual/lv_video.cpp
```diff
@@ -292,6 +293,110 @@ namespace LV {
...
+
+      return true;
+  }
+  #endif // defined(LV_ENABLE_TESTS)
+
+  bool Video::has_same_content (VideoConstPtr const& video) const
+  {
+      // Comparison of GL LV::Video is not supported.
+      if (m_impl->depth == VISUAL_VIDEO_DEPTH_GL || video->m_impl->depth == VISUAL_VIDEO_DEPTH_GL) {
+          return false;
```
hartwork, line 371: Unless we wand to avoid exceptions, this would be a case where an exception could be argued a better fit, because the two videos could still be identical and the caller maybe violated the constraints for calling this function. Just my 2 cents.

libvisual/libvisual/lv_video.cpp
```diff
@@ -292,6 +293,110 @@ namespace LV {
...
+
+      if (m_impl->depth != video->m_impl->depth)
+          return false;
+
+      // Videos must have the same colours at every pixel.
+
+      // Determine the block size in bytes in each row to compare. This must exclude padding. It must also exclude
+      // pixels outside of a subvideo's extents.
+      std::size_t const content_bytes_per_row = m_impl->width * m_impl->bpp;
```
hartwork, line 389: If we can, we should get rid of name `bpp` as quickly as we can. It's never clear whether it's bytes or bits. I hope it's bytes here?

hartwork (commented):
@kaixiong a few comments on details below. Regarding the big picture, I'm wondering if adding the testing code and using runtime logging is a bug or a feature. Is the key driver to be able to access internals easily or is there something else? Normally I would have expected to have the test code go to dedicated files and to use test-specific "assertions" (e.g. thinking gtest) rather than libvisual logging. What do you think?

libvisual/libvisual/lv_video.cpp
kaixiong: You're correct, `bpp` here means bytes per pixel.

The best part is that `visual_video_depth_bpp()` gives bits per pixel :rofl:

I'm more than ready to purge the term 'bpp' from the code.

hartwork: @kaixiong cool! :+1:

libvisual/libvisual/lv_video.cpp
kaixiong: @hartwork

> @kaixiong a few comments on details below. Regarding the big picture, I'm wondering if adding the testing code and using runtime logging is a bug or a feature. Is the key driver to be able to access internals easily or is there something else? Normally I would have expected to have the test code go to dedicated files and to use test-specific "assertions" (e.g. thinking gtest) rather than libvisual logging. What do you think?

In general, I agree.

`LV::Video::has_same_content()` is potentially useful by apps, so I made it available.

`LV::Video::validate()` is purely for internal use and testing. The function needs access to internals, and so cannot be moved to the testing code. It also comes in handy within Core itself during debugging.

We could make `validate()` part of the API but that seems wrong because `validate()` checks class invariants that will never be violated with the eventual API (where mutability of attributes is removed).

The logging is a workaround at the moment because to properly return the failed test result(s) from the function, we would need a sum/variant type (think Rust-style error handling), something I don't want to introduce just yet.

As I'm typing this comment, I realise there's a flaw in the build configuration. The examples and tools are also able to access `validate()`, so `LV_ENABLE_TESTS` has to be defined in only tests/ and libvisual/.

libvisual/libvisual/lv_video.cpp
kaixiong: @hartwork

You're right, it's neither true nor false. I'm thinking if we should throw an exception but I've been a bit hesitant to introduce exceptions into Core because it 'infects' client C++ code (like run-time type identification aka RTTI). There are platform targets where users don't want C++ exceptions built at all because of runtime costs and executable sizes. Having C++ exceptions would make life a lot easier of course. It's also consistent with how the C++ API leans heavily on construction = initialization.

Maybe it's less applicable now, I'm not sure. Anyhow, it was so controversial as exception support violated the C++'s zero-cost abstraction promise. Compilers still offer flags to turn it off together with RTTI. I believe even the C++ technical report on performance was obliged to cover it.

I continue to think about it but don't want to have to resolve this project-wide issue here for this PR.

libvisual/libvisual/lv_video.cpp
kaixiong: I agree, a direct calculation is better for validation.

libvisual/libvisual/lv_video.cpp
hartwork: > I continue to think about it but don't want to have to resolve this project-wide issue here for this PR.

@kaixiong :+1:

PR #237, opened 2023-02-02
[0.4.x] libvisual + libvisual-plugins: Fix compilation for Debian "bullseye"

hartwork:
CC @eribertomota

### DIFF
--- a/libvisual-plugins/configure.ac
+++ b/libvisual-plugins/configure.ac
@@ -51,7 +51,6 @@ dnl 'Standard' header files.
 AC_CHECK_HEADERS([fcntl.h unistd.h stdlib.h string.h sys/time.h])
 
 dnl Library functions.
-AC_FUNC_MALLOC
 AC_CHECK_FUNCS([floor gettimeofday memmove memset pow sqrt])
 AC_CHECK_FUNC([munmap], lv_have_munmap=yes, lv_have_munmap=no)
 AC_FUNC_MMAP
--- a/libvisual-plugins/configure.ac
+++ b/libvisual-plugins/configure.ac
@@ -51,7 +51,7 @@ dnl 'Standard' header files.
 AC_CHECK_HEADERS([fcntl.h unistd.h stdlib.h string.h sys/time.h])
 
 dnl Library functions.
-AC_CHECK_FUNCS([floor gettimeofday memmove memset pow sqrt])
+AC_CHECK_FUNCS([gettimeofday])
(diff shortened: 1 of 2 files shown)


libvisual-plugins/configure.ac
```diff
@@ -51,8 +51,7 @@ dnl 'Standard' header files.
 AC_CHECK_HEADERS([fcntl.h unistd.h stdlib.h string.h sys/time.h])
 
 dnl Library functions.
-AC_FUNC_MALLOC
-AC_CHECK_FUNCS([floor gettimeofday memmove memset pow sqrt])
+AC_CHECK_FUNCS([malloc floor gettimeofday memmove memset pow sqrt])
```
kaixiong, line 54: We can get rid of these checks except for `gettimeofday()`. The rest are part of the C99 standard.

hartwork: I have reverted addition of malloc and realloc to these lists just now. When checking the C99 draft linked from https://stackoverflow.com/questions/4883212/html-version-of-c-standard-draft-n1256-pdf there are more of these that I find not mentioned in there. sqrt and pow is libm (or at least can be). I'm not sure what would be the best approach here. I guess our key mission here is to catch later complete errors at configure stage. I can drop some entries but then it should be correct down to the last bit. What do you think?

hartwork: PS: @kaixiong I was replying here for the lists in both halves, if that matters. E.g. `usleep` is one of those not found but it's not up here bit in core libvisual.

hartwork: @kaixiong I have dropped all names that I could find in the C99 PDF now.

kaixiong: @hartwork thanks!

As an alternative to combing through the C standard documents, I look at manpages e.g. `man strdup` . They contain a section named 'Conforming To' that lists down all the standards (C, POSIX, etc.) the function is included in. It also mentions if the function is a compiler-specific extension. Quite a bit more informative than the C standard.

libm is a glibc specific thing, probably separated out to improve link times. It's not even necessary when building C++ programs against libstdc++, libm is indirectly linked. The C/C++ libraries I've seen on other platforms do not do this.

So when we're using glibc and expect C99 conformance, then libm has to include all of mathematics functions from the standard.

kaixiong (approved):

libvisual-plugins/configure.ac
hartwork: @kaixiong I was using man pages initially, but overlooked their `STANDARDS` section — good to know! :+1:

PR #239, opened 2023-02-02
[0.4.x] lv-tool: Increase initial/default resolution from 320x200 to 640x480

hartwork:
@kaixiong it is meant to be a conservative increase and be smaller than a quarter of 1920x1080. The idea is to just show a bit more: 320x200 is just crazy tiny these days.


kaixiong (approved):
Good call. I suppose CPUs from the last 10 years can handle the 4.8x pixel count just fine at 60Hz.

PR #241, opened 2023-02-02
libvisual-plugins: Migrate PulseAudio input plugin to PulseAudio's Asynchronous API (fixes #33)

### LINKED ISSUE #33: LV: Visualization lags when using PulseAudio

hartwork:
Fixes #33 (at least if considered to be about PulseAudio only) and hence is considered a bugfix.

This is the forward-port of 0.4.x pull request #149 to `master`.


kaixiong (approved):

kaixiong:
@hartwork, ah I've forgotten that we can straight up use [C11 atomics](https://en.cppreference.com/w/c/atomic).

hartwork:
@kaixiong let's make that a new pull request. Would you mind taking that over?

kaixiong:
> @kaixiong let's make that a new pull request. Would you mind taking that over?

@hartwork, yes I can. It's a minor change since the compiler extensions used mimicked C++11's atomics, which C11 follows closely.

PR #248, opened 2023-02-03
Plugins (pulseaudio): Use C11 atomics in place of GCC/Clang extensions.

kaixiong:
Title says it all.


hartwork (approved):
@kaixiong I checked this against https://en.cppreference.com/w/c/atomic and it looks perfect to me :+1:

PR #249, opened 2023-02-04
[0.4.x] lv-tool: Debounce resize requests (fixes #161)

### LINKED ISSUE #161: Erratic resizing of lv-tool window under GNOME Shell
When I try to resize the `lv-tool` visualization window on my GNOME Wayland desktop, the window jumps around erratically and resizes itself like it's suffering from a seizure.

hartwork:
Fixes #161

Started out at https://github.com/hartwork/sdl_video_demo/pull/10

### DIFF
--- a/libvisual/tools/lv-tool/lv-tool.cpp
+++ b/libvisual/tools/lv-tool/lv-tool.cpp
@@ -128,8 +128,8 @@ namespace LV {
             return Time(SDL_GetTicks());
         }
 
-        Time operator-(Time & other) const {
-            return Time(this->m_moment - other.m_moment);
+        friend Time operator-(Time const& lhs, Time const& rhs) {
+            return Time(lhs.m_moment - rhs.m_moment);
         }
 
         uint64_t to_usecs() const {
--- a/libvisual/tools/lv-tool/lv-tool.cpp
+++ b/libvisual/tools/lv-tool/lv-tool.cpp
@@ -132,6 +132,17 @@ namespace LV {
             return Time(lhs.m_moment - rhs.m_moment);
         }
 
+        friend bool operator>=(Time const& lhs, Time const& rhs) {
+            return lhs.m_moment >= rhs.m_moment;
+        }
+
+        //! Converts the time to seconds
+        double to_secs() const
+        {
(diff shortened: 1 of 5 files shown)


libvisual/tools/lv-tool/lv-tool.cpp
```diff
@@ -128,10 +128,14 @@ namespace LV {
             return Time(SDL_GetTicks());
         }
 
-        Time operator-(Time & other) const {
+        Time operator-(Time const& other) const {
             return Time(this->m_moment - other.m_moment);
         }
```
kaixiong, line 134: What was the reason to implement the subtraction operator as a member function? The  implementation in master uses a free function, following the practice of the C++ standard library.

libvisual/tools/lv-tool/lv-tool.cpp
```diff
@@ -455,6 +459,39 @@ namespace {
...
+
+      unsigned int width() const { return m_width; }
+      unsigned int height() const { return m_height; }
+
+      bool is_due() const {
+          if (m_applied_at >= m_received_at) {
+              return false;
+          }
+          const double seconds_passed = (LV::Time::now() - m_received_at).to_usecs() / 1000.0 / 1000.0;
+          return seconds_passed >= 0.2; // >=300ms is known to feel slow to humans
```
kaixiong, line 481: It would be better to backport `LV::Time::to_secs()` and use that instead.

libvisual/tools/lv-tool/lv-tool.cpp
hartwork: @kaixiong no particular reason, changed now as suggested.

libvisual/tools/lv-tool/lv-tool.cpp
hartwork: @kaixiong good idea, done

kaixiong (approved):
There are a few formatting inconsistencies against lv-tool in `master` which I think could be resolved to ease porting between the branches. Another enhancement  to consider is having a configurable debounce period.

Other than the above, I think the implementation looks good and could be merged first.

PR #258, opened 2023-02-06
Core (Tests): Add unit test for image loading

kaixiong:


hartwork (commented):
@kaixiong I like this pull request but it seems to have some not-terminating or infinite-loop problem: Multiple CI tasks are stuck during tests execution. Any ideas?

hartwork:
@kaixiong update: I figured now that that's what pull requests #259 is fixing. Let's merge #259 then and rebase #258 onto latest `master` and then CI should be all green with some luck.

hartwork:
PS: I have canceled the related CI runs for now so that runners can pick up other jobs again.

hartwork:
> @hartwork, it's fixed!

@kaixiong it's rebased on 7d536d1e74e0e28b9d0f4c9194672bc49f8a921d by now, I see. Nice, I won't cancel it again then, good to know! :smile: 

> There are still tests for 8-bit and 16-bit images to come. Once this PR is completed and merged, the functions developed here will be used to validate the PNG save function (#206) by means of round-tripping.

Sounds good! :+1:

hartwork (approved):
@kaixiong I found one thing:

Commit "Core (Tests): Add test for BMP image loading." seems to be doing to separate things:
- stop `load_raw_indexed_image` from converting the image to `VISUAL_VIDEO_DEPTH_24BIT`
- add BMP loading tests on top

I believe splitting this in two would yield two almost-trivial-to-review commits for the developers after you and me (and future versions of ourselves) while currently the splitting needs to be done in the reviewer head and is "actual work" and takes while.  What do you think?

kaixiong:
@hartwork Oh dear,  it looks like I missed that change in `load_raw_indexed_image` when cleaning up the commits. I agree with you, let me fix that.

kaixiong:
@hartwork I did some touching up and divided the change into two distinct commits for simplicity, one for PNG and one for BMP.

hartwork (approved):

PR #259, opened 2023-02-06
Core (LV::Video): Fix incorrect pointer increments in ARGB32 video byte swaps

kaixiong:


hartwork (approved):
@kaixiong looks good to me! :+1:

Issue #246, opened 2023-02-03
macOS support / need your help with trying 0.4.x out on macOS (branch "0.4.x-macos")


hartwork:
Hi! :wave: 

The CI covers macOS for 0.4.x for two weeks now, there is an input plugin for PortAudio that supports macOS and there have been build system fixes for macOS to make CI integration possible in the first place.

However I have no access to macOS, all my boxes run Linux (and a pile of GNU). I am looking for people to help answer these macOS questions regarding libvisual 0.4.x:
1. Does it compile out of the box?
2. If it doesn't compile, did `./configure` complain and helped figuring out what was needed?
3. Does lv-tool-0.4 start and show a spinning 3D GL spectrum?
4. Does lv-tool-0.4 use actual audio input when run with `--input portaudio` (and does it use the microphone or the audio that other apps are playing)?

If you can help answer these questions from practical experience, that would rock.
To make that easier for you here's the quickest way I know:

### Clone the right branch
```bash
git clone --branch 0.4.x-macos https://github.com/Libvisual/libvisual
cd libvisual
```

### Install homebrew packages
EDIT 2023-02-08: Add package `autoconf-archive` this list of packages to install locally:
https://github.com/Libvisual/libvisual/blob/fc109d84c613aff5dce4436f01ccc523edfd6304/.github/workflows/linux_and_macos.yml#L95-L112

### Compile and run
Here's what I would put in a script `run-0.4-macos.sh`, this is very close to what I do on Linux, locally:

```bash
#! /usr/bin/env bash
set -eux

LV_INSTALL_PREFIX="${PWD}"/INSTALL_PREFIX

configure_args=(
  --prefix="${LV_INSTALL_PREFIX}"

  # In case you would like to play with AddressSanitizer:
  #CFLAGS="-O1 -pipe -g -fsanitize=address -fno-omit-frame-pointer"
  #CXXFLAGS="-O1 -pipe -g -fsanitize=address -fno-omit-frame-pointer"
  #LDFLAGS="-g -fsanitize=address"
)

for i in libvisual libvisual-plugins ; do
  ( cd "$i" && NOCONFIGURE=please ./autogen.sh )
done

pushd libvisual
  ./configure "${configure_args[@]}"
  make -j$(nproc)
  make install
popd

pushd libvisual-plugins
  PKG_CONFIG_PATH="${LV_INSTALL_PREFIX}"/lib/pkgconfig/ \
    ./configure "${configure_args[@]}"
[...]

PR #262, opened 2023-02-07
[0.4.x] libvisual + libvisual-plugins: Cover more compilers for macOS

hartwork:
Related to #246

Was including…

- #214
- #276
- #277 
- #278
- #279
- #290

… previously.

### DIFF
--- a/.github/workflows/linux_and_macos.yml
+++ b/.github/workflows/linux_and_macos.yml
@@ -41,11 +41,36 @@ jobs:
             clang_repo_suffix:
             runs-on: ubuntu-latest
         # macOS
+          - cc: gcc-9
+            cxx: g++-9
+            clang_major_version: null
+            clang_repo_suffix: null
+            runs-on: macos-11
+          - cc: gcc-10
+            cxx: g++-10
+            clang_major_version: null
+            clang_repo_suffix: null
+            runs-on: macos-11
           - cc: gcc-11
             cxx: g++-11
             clang_major_version: null
             clang_repo_suffix: null
-            runs-on: macos-latest
+            runs-on: macos-12
+          - cc: gcc-12
+            cxx: g++-12
+            clang_major_version: null
+            clang_repo_suffix: null
+            runs-on: macos-12
+          - cc: clang
+            cxx: clang++
+            clang_major_version: null
+            clang_repo_suffix: null
+            runs-on: macos-11
+          - cc: clang
+            cxx: clang++
+            clang_major_version: null
+            clang_repo_suffix: null
+            runs-on: macos-12
     steps:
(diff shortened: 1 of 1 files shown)


hartwork:
@kaixiong rebased onto latest `0.4.x` with #214 merged now, ready for review. I can extract smaller pull requests from this and rebase again after their merge, if you see value in that.

hartwork:
@kaixiong rebased onto latest `0.4.x` now, down to 2 commits.

hartwork:
@kaixiong should I extract the last G-Force piece here to a dedicated pull request? Anything else in need of adjustment?

kaixiong:
> @kaixiong should I extract the last G-Force piece here to a dedicated pull request? Anything else in need of adjustment?

@hartwork That would be good, thanks.

hartwork:
> > @kaixiong should I extract the last G-Force piece here to a dedicated pull request? Anything else in need of adjustment?
> 
> @hartwork That would be good, thanks.

@kaixiong here you go: #290

hartwork:
@kaixiong now rebased off #290 onto latest `0.4.x`.

kaixiong:
@hartwork Digressing a bit but do we need to cover this many compiler versions? I welcome the idea but I feel that the time spent on CI is not scaling well.

hartwork:
@kaixiong do you mean scaling well as in human-perceived CI runtime or as in total machine resource usage?

On other projects (with Python) where I want to be covered against surprises with Python versions in 3.7, 3.8, 3.9, 3.10, 3.11 I often cover the oldest supported and the most recent one but not the intermediates, i.e. 3.7 and 3.11 in that case, and that works okay. However our situation in Libvisual is that we have multiple macOS versions (more dimensions) + developers not using macOS (extra blindness and value) + different past errors than found by Linux CI (also extra value). We can trim this down more but I wonder at what loss and what value.

What do you think?

hartwork:
@kaixiong PS: Three more tiny notes:
- I have not added/used `macos-10.15` anywhere because it's already deprecated (in the eyes of https://github.com/actions/runner-images)
- (This is already far from a complete coverage of all combinations.)
- The matrix entries I'd be least reluctant to give up are GCC 10 and GCC 11 (because we'll still have GCC 9@macos-11 and GCC 12@macos-12).

hartwork:
@kaixiong what do you think?

kaixiong:
@hartwork, I think the additional build actions for MacOS are okay.

I was mostly concerned if we're going to do the same for other platforms (OS+compiler combos). That would have a multiplicative effect. If we start adding other kinds of validation (UBSAN, ASAN) in the future, it's going to take even longer.

hartwork:
@kaixiong cool! :+1:

PR #265, opened 2023-02-08
Core: Add LV_NODISCARD for C API functions with non-discardable results  

kaixiong:
This PR adds the define `LV_NODISCARD` for marking C API functions whose return results should not be discarded. I have added them to API functions that return allocated objects.

The define uses compiler-specific function attributes to achieve this when building C code.

For C++, `[[nodiscard]]` already exists from C++17 onwards. C++20 has a variant that accepts a string literal for explaining the rationale that could be displayed in compiler messages. I can't see how to achieve the same for C unfortunately.

Use `LV_NODISCARD` only for C API functions. `[[nodiscard]]` for C++ functions.

### DIFF
--- a/libvisual/libvisual/lv_defines.h
+++ b/libvisual/libvisual/lv_defines.h
@@ -120,7 +120,9 @@
   #define __PRETTY_FUNCTION__ __FUNCTION__
 #endif
 
-#ifdef _MSC_VER
+#ifdef __cplusplus
+  #define LV_NODISCARD [[nodiscard]]
+#elif defined(_MSC_VER)
   #define LV_NODISCARD _Check_return_
 #elif defined(__GNUC__)
   #define LV_NODISCARD  __attribute__ ((warn_unused_result))
--- a/libvisual/libvisual/lv_defines.h
+++ b/libvisual/libvisual/lv_defines.h
@@ -120,7 +120,7 @@
   #define __PRETTY_FUNCTION__ __FUNCTION__
 #endif
 
-#ifdef __cplusplus
+#if __cplusplus >= 201703L
   #define LV_NODISCARD [[nodiscard]]
 #elif defined(_MSC_VER) && _MSC_VER >= 1700
   #define LV_NODISCARD _Check_return_
--- a/libvisual/libvisual/lv_defines.h
+++ b/libvisual/libvisual/lv_defines.h
@@ -115,8 +115,17 @@
 #define LV_PLUGIN_EXPORT LV_DLL_EXPORT
 
 /* Utility macros */
+
 #ifdef _MSC_VER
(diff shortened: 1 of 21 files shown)


hartwork (commented):
@kaixiong I like the idea. I think I found three cases that slipped through the cracks:

```console
# git grep 'LV_API.*_new' | grep -v LV_NODISCARD
libvisual/libvisual/lv_param_value.h:LV_API VisParamValue *visual_param_value_new        (VisParamType   type, void *value);
libvisual/libvisual/lv_time.h:LV_API VisTimer *visual_timer_new  (void);

# git grep 'LV_API.*clone' | grep -v LV_NODISCARD
libvisual/libvisual/lv_palette.h:LV_API VisPalette *visual_palette_clone (VisPalette *self);
```
What do you think?

libvisual/libvisual/lv_defines.h
```diff
@@ -115,8 +115,19 @@
 #define LV_PLUGIN_EXPORT LV_DLL_EXPORT
 
 /* Utility macros */
+
 #ifdef _MSC_VER
   #define __PRETTY_FUNCTION__ __FUNCTION__
 #endif
 
+#if __cplusplus >= 201703L
```
hartwork, line 123: @kaixiong could it be that this needs addition of `defined(__cplusplus) && ` ?

hartwork: @kaixiong let me first spell out the questions that I think we need to answer and then try to give a potential answer. For the questions:
1. Is this code used with a C compiler or only C++? Does need it to work from plain C?
2. Can we safely compare an undefined macro against an integer value?
3. Does the `__cplusplus` value reflect reality (more below)?

Potential answers:
1. Seems like yes, we need to support both C and C++ here.
2. It is defined according to https://stackoverflow.com/questions/5085392/what-is-the-value-of-an-undefined-constant-used-in-if so it seems to become a question of style and/or clarity.
3. MSVC [is said](https://bloomfield.online/posts/report-__cplusplus-correctly-with-cmake-and-visual-studio/) to not have reality in `__cplusplus`. The post lays out some options. [Another one](https://developercommunity.visualstudio.com/t/msvc-incorrectly-defines-cplusplus/139261). I think we do want to support MSVC?

PS: https://sourceforge.net/p/predef/wiki/Standards/ does not seem to have any additional insights.

What do you think?

kaixiong: @hartwork Looks like Question 1 and 2 are effectively the same question as the C/C++ preprocessor both default to 0 for undefined constants (apart from the handling of `true` and `false`).

Question 3 is addressed in #269 but does not ultimately matter for this PR because the code falls back to compiler extensions (`_Check_return_`) immediately below this.

hartwork (approved):

PR #266, opened 2023-02-08
Core: Fix wrong define used to check for GCC version.

kaixiong:


hartwork (approved):

kaixiong:
@hartwork, the branch name is wrong. Is there a way to change it while preserving the PR?

hartwork:
@kaixiong I'm afraid there is not, no.

kaixiong:
@hartwork I'll have to bite the bullet and redo this with a new PR then :facepalm:

PR #269, opened 2023-02-10
Core (Build): Get MSVC to use correct values for the __cplusplus define

kaixiong:
Rant: As usual, Microsoft does its own thing.

References:
- Documentation on Microsoft Learn ([link](https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170))
- Thread on CMake ([link](https://gitlab.kitware.com/cmake/cmake/-/issues/18837))

### DIFF
--- a/libvisual/CMakeLists.txt
+++ b/libvisual/CMakeLists.txt
@@ -51,6 +51,12 @@ SET(CMAKE_CXX_STANDARD 20)
 SET(CMAKE_CXX_STANDARD_REQUIRED on)
 SET(CMAKE_CXX_EXTENSIONS off)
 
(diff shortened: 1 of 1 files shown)


hartwork (approved):

hartwork:
@kaixiong we can make the CI cover MSVC at some point.

kaixiong:
> @kaixiong PS: On a side note we're saying goodbye to Visual Studio <2017 (at least in part) here. Fine in my book, but important to know.

`@hartwork` We already are, `master `requires C++20 to build.

hartwork:
@kaixiong I guess you're right  — https://en.cppreference.com/w/cpp/compiler_support#cpp20

PR #271, opened 2023-02-10
Core (lv-tool): Protect new display driver instances using unique_ptr

kaixiong:
`DisplayDriverFactory::make()` currently returns driver instances in raw pointers. This patch changes it to wrap them in `std::unique_ptr`.

Note that the various constructors `stdout_driver_new()`, `stdout_sdl_driver_new()`, etc. still return raw pointers. The driver code are meant to be externalised into loadable SOs and I'm unsure if we can return `std::unique_ptr` from there.

### DIFF
--- a/libvisual/tools/lv-tool/display/display.cpp
+++ b/libvisual/tools/lv-tool/display/display.cpp
@@ -25,6 +25,7 @@
 #include "display.hpp"
 #include "display_driver_factory.hpp"
 #include <libvisual/libvisual.h>
+#include <utility>
 #include <stdexcept>
 #include <string>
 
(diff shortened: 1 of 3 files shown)


libvisual/tools/lv-tool/display/display.cpp
```diff
@@ -34,19 +35,17 @@ class Display::Impl
...
+        : driver {std::move (driver_)}
     {}
 
     ~Impl ()
     {}
 };
 
 Display::Display (std::string const& driver_name)
-  : m_impl (new Impl)
+    : m_impl {new Impl {DisplayDriverFactory::instance().make (driver_name, *this)}}
```
hartwork, line 47: @kaixiong we are passing `*this` here while `this` is in the middle of construction and not ready for any use beyond maybe holding a reference or pointer to it. We are sort-of doing that already before but… does we even need that? If it's out of scope, can it be resolved later?

kaixiong: @hartwork good catch! I forgot about this but it's definitely safe with the current code since every driver class only stores the reference during their construction. But let me think about it more.

PR #276, opened 2023-02-13
[0.4.x] libvisual + libvisual-plugins: Require a C++11 compiler

hartwork:


kaixiong:
@hartwork `master` avoids `reinterpret_cast` by specifying the 4-character value in hexadecimals or using a macro to build up the 4-byte integer value instead. It'd be good to follow it. On the other hand, the changes are quite large and were also made to make it work on Windows (I think?). I'll leave it up to you to decide.

References: https://github.com/Libvisual/libvisual/commit/f59364b840b136dae3dcb5a47e6075cbb0bcc561, https://github.com/Libvisual/libvisual/commit/e012c97c2b9f6e2646da98203444279bb99767c4

kaixiong (approved):

hartwork:
@kaixiong maybe the `VAL(a,b,c,d)` approach over there is best, e.g. because it self-answers all potential questions about endianness. Let's first get the compilation fixed and then use `VAL(a,b,c,d)` eventually, clean up more and being more cross-branch-consistent. Thank you! :+1:

PR #277, opened 2023-02-13
[0.4.x] libvisual-plugins: Resolve use of function "mremap" in input plug-in "mplayer" (e.g. for macOS)

hartwork:


libvisual-plugins/plugins/input/mplayer/input_mplayer.c
```diff
@@ -180,9 +178,12 @@ int inp_mplayer_init( VisPluginData *plugin )
...
 
-	priv->mmap_area = mremap( priv->mmap_area, sizeof( mplayer_data_t ),
-			sizeof( mplayer_data_t ) + priv->mmap_area->bs,
-			0 );
+	const int buffer_size = priv->mmap_area->bs;
+	munmap( priv->mmap_area, sizeof( mplayer_data_t ) );
+	priv->mmap_area = NULL;
+
+	priv->mmap_area = mmap( 0, sizeof( mplayer_data_t ) + buffer_size,
+			PROT_READ, MAP_SHARED, priv->fd, 0 );
```
kaixiong, line 186: Is there a need to set `mmap_area` first to `NULL`? It looks like `mmap()` is guaranteed to set `mmap_area` regardless of success or error.

hartwork: @kaixiong you're right, fixed and pushed.

kaixiong (approved):

PR #278, opened 2023-02-13
[0.4.x] libvisual-plugins: Compile and link against OpenGL and GLU in a more portable way (e.g. for macOS)

hartwork:


kaixiong (commented):
@hartwork, GL actors should be excluded from build if the required OpenGL headers are not found.

hartwork:
@kaixiong we have at least some of that:
```
# git grep HAVE_OPENGL | grep test
libvisual-plugins/configure.ac:  AS_IF([test "$HAVE_OPENGL" != "yes"], [
libvisual-plugins/configure.ac:  if test "$HAVE_OPENGL" = "yes"; then
libvisual-plugins/configure.ac:  if test "$HAVE_OPENGL" = "yes"; then
libvisual-plugins/configure.ac:  if test "$HAVE_OPENGL" = "yes"; then
libvisual-plugins/configure.ac:  if test "$HAVE_OPENGL" = "yes"; then
```
If we need, more happy to do more in a new and dedicated pull request.

kaixiong:
@hartwork, Ah okay. If that's the case, then we do not need the '#error' when including GL/GLU headers. But I guess it doesn't really matter.

kaixiong (approved):

hartwork:
@kaixiong the `#error` was was made me notice that `config.h` was note included in two places. It would have failed later anyway, but my vote for keeping it, it's like a better compile-time assertion.

kaixiong:
> @kaixiong the `#error` was was made me notice that `config.h` was not included in two places. It would have failed later anyway, but my vote for keeping it, it's like a better compile-time assertion.

@hartwork I think my initial annoyance was more to do with having to complicate the header inclusion logic due to MacOS's unorthodox location. I'm fine with `#error` either way. I hope that when we do use GLEW or GLAD, we can kill this abomination with a single  `#include`.

hartwork:
@kaixiong we should double check that GLEW pulls in OpenGL via `-framework` and we should only drop the direct GL includes if we no longer call any GL functions directly. Then...

PR #279, opened 2023-02-13
[0.4.x] libvisual-plugins: Make input plugin "portaudio" fall back to mono as needed (e.g. for macOS)

hartwork:


libvisual-plugins/plugins/input/portaudio/input_portaudio.c
```diff
@@ -39,6 +37,7 @@ VISUAL_PLUGIN_API_VERSION_VALIDATOR
 
 typedef struct {
 	PaStream *stream;
+	unsigned count_channels;  // 0 (unknown), 1 (mono) or 2 (stereo)
```
kaixiong, line 40: Would it be better to name this `channel_count` and use `unsigned int` (instead of `unsigned`)?

hartwork: @kaixiong can do. `unsigned` on its own is `unsigned int`, it was not by accident. Done, pushed.

kaixiong (approved):

PR #284, opened 2023-02-18
Core (Memory): Remove custom implementations of memset() and memcpy().

kaixiong:
This closes #275.

PR #285, opened 2023-02-18
Core (Memory): Re-implement visual_mem_set{16,32} to exploit GCC/Clang auto-vectorisation

kaixiong:
Builds on PR #284.

### DIFF
--- a/libvisual/tests/mem_test/mem_set_test.cpp
+++ b/libvisual/tests/mem_test/mem_set_test.cpp
@@ -1,0 +1,68 @@
+#include "test.h"
+#include <cstdint>
+#include <libvisual/libvisual.h>
+#include <libvisual/lv_mem.h>
+#include <cstdint>
+#include <array>
+#include <algorithm>
+#include <vector>
+
+namespace {
+
+  void test_mem_set16 ()
+  {
+      constexpr std::uint16_t test_pattern = 0x1234;
+
+      constexpr auto test {[test_pattern] (auto x) {
+                               return x == test_pattern;
+                           }};
+
+      for (auto test_size = 1; test_size <= 128; test_size++) {
+          // Allocate extra elements in front and behind of test memory area.
+          std::vector<std::uint16_t> vec (test_size + 2);
+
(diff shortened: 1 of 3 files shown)


libvisual/tests/mem_test/mem_set_test.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+      constexpr auto test {[test_pattern] (auto x) {
+                               return x == test_pattern;
+                           }};
+
+      for (auto test_size = 1; test_size <= 128; test_size++) {
+          // Allocate extra elements in front and behind of test memory area.
+          std::vector<std::uint32_t> vec (test_size + 2);
+
+          visual_mem_set32 (&vec[1], test_pattern, test_size);
```
hartwork, line 47: @kaixiong I'd expect `test_size * sizeof(std::uint32_t)` instead of plain `test_size` here — I am missing something? One more case like that below.

hartwork: @kaixiong nevermind, I just figured what the `32` in the function name is for...

hartwork (approved):
@kaixiong I cannot offer proper review of the intrinsics part as of today, but the tests look good, they pass, and I trust that you made this with care :+1:

kaixiong:
@hartwork After some testing, I've found that rewriting the Standard C versions to exploit GCC and Clang's autovectorisation at `-O2` is more portable and performant than the MMX/SSE SIMD implementation for buffers that are 16MB or less in size*. That is the size of a 2048x2048 32-bit image. This should be good enough for our purposes for now.

* Any larger, we would have to employ x86 SIMD intrinsics again to use non-temporal SSE2 instructions.

kaixiong:
Added SSE2 implementations using `_mm_storeu_si128` to make up for the lack of auto-vectorisation in GCC versions 11.x and older at `-O2`. The implementation is simple and suffers a bit at buffer sizes <= 128 compared to the auto-vectorised standard C versions.

Benchmark code can be found [here](https://gist.github.com/kaixiong/89607caddf7f84c5cb45b4fdd10900be).

PR #286, opened 2023-02-18
Core (Build): Fix LV .so compile and link options

kaixiong:
This PR consists of two patches.

1. First patch fixes the way compile and link options are set for the core .so. This brings back the compiler flags such `-Wall` set at the top level.

2. Second patch removes `--no-undefined` (equivalent of `-z defs`) from the .so linker options. This fixes the unresolved symbol errors when Clang sanitizers are enabled and is recommended by the official Clang ASAN documentation.

@hartwork I'm unsure of change #2. `--no-undefined` can help check missing symbols early. Should we instead  remove the option only when building with Clang and `-fsanitize=xxx` is included in the build flags?


hartwork (commented):
> 1. First patch fixes the way compile and link options are set for the core .so. This brings back the compiler flags such `-Wall` set at the top level.

@kaixiong interesting catch, cool!

> 2. Second patch removes `--no-undefined` (equivalent of `-z defs`) from the .so linker options. This fixes the unresolved symbol errors when Clang sanitizers are enabled and is recommended by the official Clang ASAN documentation.
> 
> @hartwork I'm unsure of change #2. `--no-undefined` can help check missing symbols early. Should we instead remove the option only when building with Clang and `-fsanitize=xxx` is included in the build flags?

I'm not fully sure right now how `.o` files in isolation would be checked against missing symbols when they are not linked against libraries yet :thinking: What is your understanding of how `--no-undefined` operates? I understand `--no-allow-shlib-undefined` but `--no-undefined`?
Getting rid of `--unresolved-symbols,ignore-in-shared-libs` seems like a good move.
Regarding Clang and sanitizers, we have a different approach to this problem on `0.4.x` from 
b18dd14746343150dad963a50ee13156b8ce099e that may work for both Clang and non-Clang (but I'm open to whatever is waterproof).

What do you think?

hartwork:
@kaixiong PS: For the CI error, I believe we have to drop the two `LV_RESTRICT`s from the function signature of `visual_math_simd_mul_floats_float`. What do you think?

kaixiong:
> @kaixiong PS: For the CI error, I believe we have to drop the two `LV_RESTRICT`s from the function signature of `visual_math_simd_mul_floats_float`. What do you think? Is `simd_mul_floats_float` okay with source and destination being identical?

@hartwork I don't really know how to view ORC's generated SIMD instructions but its C wrapper functions do have the same restrict attributes applied. I fixed this in `master` for now by using intermediate buffers and rebased this branch. It builds fine on CI now.

P.S. In a hurry, I made the change _directly_ on `master` by mistake and force pushed it, so there has been no PR filed for this fix. I've added a branch protection rule for `master` to prevent this in the future.

P.P.S. The way buffers are used in `LV::Audio` looks horrendous (it's a straight port from the C code). I'd have to look at it again when I have time.

kaixiong:
> @hartwork  I'm not fully sure right now how `.o` files in isolation would be checked against missing symbols when they are not linked against libraries yet thinking What is your understanding of how `--no-undefined` operates? I understand `--no-allow-shlib-undefined` but `--no-undefined`? Getting rid of `--unresolved-symbols,ignore-in-shared-libs` seems like a good move. Regarding Clang and sanitizers, we have a different approach to this problem on `0.4.x` from [b18dd14](https://github.com/Libvisual/libvisual/commit/b18dd14746343150dad963a50ee13156b8ce099e) that may work for both Clang and non-Clang (but I'm open to whatever is waterproof).

@hartwork My understanding is that with `--no-undefined`, the linker verifies that all symbols in the SO target resolves to something, as if it were building an executable. This is off by default.

Do note that this happens at the link stage when the `.o` files are brought together, not during compilation when the compiler is invoked to create individual `.o` files from C/C++ sources.

I guess the verification is not strictly necessary as the complete symbol resolution can be performed later when a program is linked against the library SO. After all, an SO is not a complete program &mdash; some of its symbol references may well be in the executable! Or SOs could be mutually dependent (A has symbol references to B and vice versa) and as such, the external symbols can only be resolved at a later stage.
 
I believe what's going on when building lv-tool is that Clang's sanitizer libraries are statically linked into the executable. During runtime, the sanitizer symbol references in libvisual are dynamically resolved. This serves to explain why lv-tool does not have any runtime dependency on Clang's shared libraries as well.

GCC chose instead  to link in its shared library when building libvisual. In doing so, it does not incur any errors with `--no-undefined`.

Another thought: Clang probably chose to statically link its sanitizer libraries into the executable instead of the SO (which we might expect) so that the sanitizer code is centralised. Statically linking into libraries would be a mistake because a copy of the sanitizer in each linked SO would have its own state! 

Finally, Clang probably chose static linking over dynamic linking because its libraries are not often in the OS search path.

hartwork:
> @hartwork I don't really know how to view ORC's generated SIMD instructions but its C wrapper functions do have the same restrict attributes applied. I fixed this in `master` for now by using intermediate buffers and rebased this branch. It builds fine on CI now.

@kaixiong I see, cool to have this fixed.


> P.S. In a hurry, I made the change _directly_ on `master` by mistake and force pushed it, so there has been no PR filed for this fix. I've added a branch protection rule for `master` to prevent this in the future.

I understand, no worries, good move.

> @hartwork My understanding is that with `--no-undefined`, the linker verifies that all symbols in the SO target resolves to something, as if it were building an executable. This is off by default.
> 
> Do note that this happens at the link stage when the `.o` files are brought together, not during compilation when the compiler is invoked to create individual `.o` files from C/C++ sources.

Isn't that what we already have the other flags like `--no-allow-shlib-undefined` for? :thinking: 


> I guess the verification is not strictly necessary as the complete symbol resolution can be performed later when a program is linked against the library SO. After all, an SO is not a complete program — some of its symbol references may well be in the executable! SOs could also be mutually dependent (A has symbol references to B and vice versa) and as such, their external symbols can only be resolved at a later stage.

Yes, but it also makes us "vulnerable" to running into unresolved symbols at load time. if we limit ourselves to satisfying all symbols after the build, we will make sure that users have little chance to run into this problem later.

> I believe what's going on when building lv-tool is that Clang's sanitizer libraries are statically linked into the executable. During runtime, the sanitizer symbol references in libvisual are dynamically resolved. This serves to explain why lv-tool does not have any runtime dependency on Clang's shared libraries as well.
> 
> GCC chose instead to link in its shared library when building libvisual. In doing so, It does not incur any errors with `--no-undefined`.
> 
> Another thought: Clang probably chose to statically link its sanitizer libraries into the executable instead of the SO (which we might expect) so that the sanitizer code is centralised. Statically linking into libraries would be a mistake because a copy of the sanitizer in each linked SO would have its own state!
> 
> Finally, Clang probably chose static linking over dynamic linking because its libraries are not often in the OS search path.

Let me digest this more.

We can merge in the current state if you are good with where the pull request is now?

kaixiong:
> Isn't that what we already have the other flags like `--no-allow-shlib-undefined` for? thinking

@hartwork  `--no-allow-shlib-undefined` only applies to symbols in SOs that are being _linked against_ (libpng for libvisual.so for example).

`--no-undefined` applies to `.o` files. For libvisual.so,  it's the `.o` files built with Clang instrumentation that contain references to Clang's sanitizer.

> Yes, but it also makes us "vulnerable" to running into unresolved symbols at load time. if we limit ourselves to satisfying all symbols after the build, we will make sure that users have little chance to run into this problem later.

Yup, this is the source of my hesitation with removing `--no-undefined`. On the other hand, `lv-tool` is included in the default build and triggers the validation anyway so there is still quality control.

The TLDR version is that, by default, Clang links the .a file for the relevant sanitizer into the executable (e.g. `lv-tool`), and the sanitizer symbol references in core and plugins are resolved during runtime. Adding `--no-undefined` breaks this model.

> We can merge in the current state if you are good with where the pull request is now?

If the purpose of removing `--no-undefined` and other link flags is just to have Core and plugins build properly with Clang sanitizers, I'm wondering if we could limit the 'damage' by detecting Clang and `-fsanitize=.*` in the compile flags and remove the linker checks accordingly.

hartwork:
> > Yes, but it also makes us "vulnerable" to running into unresolved symbols at load time. if we limit ourselves to satisfying all symbols after the build, we will make sure that users have little chance to run into this problem later.
> 
> Yup, this is the source of my hesitation with removing `--no-undefined`. On the other hand, `lv-tool` is included in the default build and triggers the validation anyway so there is still quality control.

@kaixiong yes, lv-tool is convered. My concern is the plug-ins.


> If the purpose of removing `--no-undefined` and other link flags is just to have Core and plugins build properly with Clang sanitizers, I'm wondering if we could limit the 'damage' by detecting Clang and `-fsanitize=.*` in the compile flags and remove the linker checks accordingly.

I think I'm happy with "any" solution that doesn't break compilation with GCC, Clang and/or MSVC and protects users from future plug-in load errors.

kaixiong:
@hartwork, I've made the unresolved symbol checks an opt-out build option. Probing compiler flags looked a bit too hairy.

hartwork:
@kaixiong can we resolve the double negation from "disable no" to "enable yes" please? The technical complexity is the same but the human-perceived complexity is way lower.

kaixiong:
@hartwork Good idea, done.

hartwork (approved):
@kaixiong thanks!

PR #295, opened 2023-03-20
[0.4.x] Trim down CI matrix a bit

hartwork:
@kaixiong these are resources we could save without too much risk, if we want. What do you think?


kaixiong:
@hartwork, I think this is okay for the C99 and C++11 code in 0.4.x.

kaixiong (approved):

Issue #300, opened 2023-03-26
version 0.4.2: ./configure: 18070: Syntax error: word unexpected (expecting ")")


TimB87:
Hi all :wave: 

While trying to build a 32 bit library of libvisual 0.4.2, I get this weird error while running `configure`:
```
=======> Building '/home/pkgmk/packages/libvisual-32#0.4.2-1.pkg.tar.xz'.
bsdtar -p -o -C /home/pkgmk/work/libvisual-32/src -xf /home/pkgmk/distfiles/libvisual-0.4.2.tar.bz2
+ build
+ cd libvisual-0.4.2
+ export PKG_CONFIG_PATH=/usr/lib32/pkgconfig
+ PKG_CONFIG_PATH=/usr/lib32/pkgconfig
+ sed -i s/sdl-config/sdl-config-32/g configure.ac
+ autoreconf -vfi
autoreconf: export WARNINGS=
autoreconf: Entering directory '.'
autoreconf: running: autopoint --force
Copying file m4/codeset.m4
Copying file m4/extern-inline.m4
Copying file m4/fcntl-o.m4
Copying file m4/glibc2.m4
Copying file m4/glibc21.m4
Copying file m4/intdiv0.m4
Copying file m4/intl.m4
Copying file m4/intldir.m4
Copying file m4/intmax.m4
Copying file m4/inttypes-pri.m4
Copying file m4/inttypes_h.m4
Copying file m4/lcmessage.m4
Copying file m4/lock.m4
Copying file m4/longlong.m4
Copying file m4/printf-posix.m4
Copying file m4/size_max.m4
Copying file m4/stdint_h.m4
Copying file m4/threadlib.m4
Copying file m4/uintmax_t.m4
Copying file m4/visibility.m4
Copying file m4/wchar_t.m4
Copying file m4/wint_t.m4
Copying file m4/xsize.m4
Copying file po/Makevars.template
autoreconf: running: aclocal --force -I m4
autoreconf: configure.ac: tracing
autoreconf: running: libtoolize --copy --force
libtoolize: putting auxiliary files in '.'.
libtoolize: copying file './ltmain.sh'
libtoolize: putting macros in 'm4'.
libtoolize: copying file 'm4/libtool.m4'
libtoolize: copying file 'm4/ltoptions.m4'
libtoolize: copying file 'm4/ltsugar.m4'
libtoolize: copying file 'm4/ltversion.m4'
libtoolize: copying file 'm4/lt~obsolete.m4'
[...]

PR #301, opened 2023-03-26
[0.4.x] libvisual: configure.ac: Resolve Bashism

hartwork:
In reaction to https://github.com/Libvisual/libvisual/issues/300#issuecomment-1484167603

### DIFF
--- a/libvisual/configure.ac
+++ b/libvisual/configure.ac
@@ -352,7 +352,7 @@ build_subdirs="libvisual"
 VISUAL_PKG_CONFIG_DEP=""
 
 VISUAL_CFLAGS=""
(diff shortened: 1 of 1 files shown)


kaixiong (approved):
Today I learned that Bash supports `==` :laughing:

hartwork:
@kaixiong you're welcome :smile:

eli-schwartz:
Having just hit this while testing libvisual-plugins for LTO / strict-aliasing issues... seems it's not in a release yet though. :)

> Today I learned that Bash supports `==` 😆

Note that bash supports it as a direct alias, it is exactly equal to writing "=" and thus serves no purpose at all other than to horribly confuse people and break muscle memory. You're better off not knowing it exists. ;)

PR #325, opened 2023-10-27
libvisual-plugins (lcdcontrol): Address Clang 18 warning `-Wvla-cxx-extension` to fix CI

hartwork:
Symptom was:
```
[..]/libvisual-plugins/plugins/actor/lcdcontrol/GenericSerial.cpp:141:15: error: variable length arrays in C++ are a Clang extension [-Werror,-Wvla-cxx-extension]
  141 |     char buff[len + 1];
      |               ^~~~~~~
```

.. plus the same a few times more.


kaixiong:
Ideally we should use `std::vector` for this but this is okay for now.

hartwork:
@kaixiong thanks!

PR #351, opened 2024-05-30
[0.4.x] libvisual: fix accidental assignment instead of equality check

triallax:
with the accidental assignment, the function would always return 0 for a keytype of VISUAL_HASHMAP_KEY_TYPE_STRING (because it has a value of 2), which is technically valid but not good for obvious reasons

### DIFF
--- a/libvisual/libvisual/lv_hashmap.c
+++ b/libvisual/libvisual/lv_hashmap.c
@@ -261,7 +261,7 @@ static int get_hash (VisHashmap *hashmap, void *key, VisHashmapKeyType keytype)
 {
 	if (keytype == VISUAL_HASHMAP_KEY_TYPE_INTEGER)
 		return integer_hash (*((uint32_t *) key)) % hashmap->tablesize;
-	else if (keytype = VISUAL_HASHMAP_KEY_TYPE_STRING)
+	else if (keytype == VISUAL_HASHMAP_KEY_TYPE_STRING)
 		return string_hash ((char *) key) % hashmap->tablesize;
 
 	return 0;


hartwork (approved):
@kaixiong what do you think?

kaixiong:
Nice find. Neither Clang nor GCC warns about this assignment?

triallax:
i don't know about gcc, but i found out about this because clang warned about it when i was bumping libvisual for chimera linux:

```
lv_hashmap.c:264:19: warning: using the result of an assignment as a condition without parentheses [-Wparentheses]
  264 |         else if (keytype = VISUAL_HASHMAP_KEY_TYPE_STRING)
      |                  ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
lv_hashmap.c:264:19: note: place parentheses around the assignment to silence this warning
  264 |         else if (keytype = VISUAL_HASHMAP_KEY_TYPE_STRING)
      |                          ^                               
      |                  (                                       )
lv_hashmap.c:264:19: note: use '==' to turn this assignment into an equality comparison
  264 |         else if (keytype = VISUAL_HASHMAP_KEY_TYPE_STRING)
      |                          ^
      |                          ==
```

kaixiong:
> @kaixiong what do you think?

@hartwork would adding `-Werror` be a good idea? I haven't had to look at or build 0.4.x in a long time to tell.

hartwork:
@kaixiong _outside_ of `configure.ac` in CI that would be great — yes! —, but my memory says that we're not at zero with warnings and so we cannot enable it without fixing more warnings first. Would need a closer look.

hartwork:
@kaixiong I inspected the `-Werror` option more and there is apparently a long road for `-Werror` in CI to go. I have created pull request #375 with some of the easier fixes, but not all of the remaining warnings are trivial to fix. Some are definitely pointing to unfixed bugs.

PR #378, opened 2025-01-15
Core (Build): Get ORC compiler executable path from pkg-config.

kaixiong:
ORC's pkg-config file conveniently contains the path to the ORC compiler (orcc), so we should use that instead of searching for the executable in `PATH`.


hartwork (approved):
@kaixiong there is no use of ORC in 0.4.x code that I could find, is that correct? If so, I would create a pull request dropping installation of ORC from 0.4.x CI.

kaixiong:
@hartwork yes, ORC was introduced in `master`.

PR #383, opened 2025-01-21
Plugins (alphablend): Fix blending and associated crashes due to buffer overruns

kaixiong:
The alphablend morph uses the blending routines `visual_alpha_blend_nn()` wrongly. It passes the pixel buffer size in bytes instead of pixel count for the `size` parameter. Since buffer size is always equal or greater than pixel count (bytes per pixel >= 1 and row padding >= 0), this causes buffer overruns.

### DIFF
--- a/libvisual-plugins/plugins/morph/alphablend/morph_alphablend.c
+++ b/libvisual-plugins/plugins/morph/alphablend/morph_alphablend.c
@@ -24,6 +24,7 @@
 #include "config.h"
 #include "gettext.h"
 #include <libvisual/libvisual.h>
+#include <stdlib.h>
 
 VISUAL_PLUGIN_API_VERSION_VALIDATOR
 
@@ -31,7 +32,9 @@ static int  lv_morph_alpha_init    (VisPluginData *plugin);
 static void lv_morph_alpha_cleanup (VisPluginData *plugin);
 static void lv_morph_alpha_apply   (VisPluginData *plugin, float progress, VisAudio *audio, VisVideo *dest, VisVideo *src1, VisVideo *src2);
 
-static inline void alpha_blend_buffer (uint8_t *dest, uint8_t *src1, uint8_t *src2, int size, int depth, float alpha);
+typedef void (*BlendFunc) (uint8_t *, const uint8_t *, const uint8_t *, visual_size_t, uint8_t);
+
+static BlendFunc get_blend_func (VisVideoDepth depth);
 
 const VisPluginInfo *get_plugin_info (void)
 {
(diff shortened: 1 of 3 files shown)


libvisual-plugins/plugins/morph/alphablend/morph_alphablend.c
```diff
@@ -80,33 +82,38 @@ static void lv_morph_alpha_cleanup (VisPluginData *plugin)
...
+	int height = visual_video_get_height (dest);
+	int depth  = visual_video_get_depth (dest);
+	int pitch  = visual_video_get_pitch (dest);
+
+	uint8_t *src1_row_ptr = visual_video_get_pixels (src1);
+	uint8_t *src2_row_ptr = visual_video_get_pixels (src2);
+	uint8_t *dest_row_ptr = visual_video_get_pixels (dest);
+
+	uint8_t alpha = progress * 255;
+	BlendFunc blend_func = get_blend_func (depth);
```
hartwork, line 95: `get_blend_func` can return `NULL`. If we keep allowing it to, we likely need a check for `NULL` somewhere around here?

libvisual-plugins/plugins/morph/alphablend/morph_alphablend.c
```diff
@@ -80,33 +82,38 @@ static void lv_morph_alpha_cleanup (VisPluginData *plugin)
...
+	int width  = visual_video_get_width (dest);
+	int height = visual_video_get_height (dest);
+	int depth  = visual_video_get_depth (dest);
+	int pitch  = visual_video_get_pitch (dest);
+
+	uint8_t *src1_row_ptr = visual_video_get_pixels (src1);
+	uint8_t *src2_row_ptr = visual_video_get_pixels (src2);
+	uint8_t *dest_row_ptr = visual_video_get_pixels (dest);
+
+	uint8_t alpha = progress * 255;
```
hartwork, line 94: The original formula was `uint8_t a = alpha * (1/255.0);` with rather different results, unless I am missing something. Was the old formula just wrong and the new formula fixes it or is this change accidental?

PS: If that that part of the change could be made a separate commit, I think it could help Git history clarity.

hartwork (changes requested):
@kaixiong two small things:

libvisual-plugins/plugins/morph/alphablend/morph_alphablend.c
kaixiong: The existing code is wrong. Across the morph transition, `alpha` should move from 0 to 255 as `progress`  is increased from 0.0 to 1.0. This increases the weight of the blended-in image as the morph progresses.

Oddly enough, 0.4.x has the correct (and equivalent) formula so I might've inverted the two ranges due to confusion.

Let me split this change out into a separate commit.

libvisual-plugins/plugins/morph/alphablend/morph_alphablend.c
kaixiong: The default case is unreachable and largely included to silence compiler warnings about enum values unaccounted for (e.g. `VISUAL_VIDEO_DEPTH_NONE` and `VISUAL_DEPTH_NONE_ALL`).

We don't have `unreachable()` in C99 or C11 so I returned a `NULL` instead. Would you prefer `abort()`?

hartwork: @kaixiong I would favor some flavor of `abort` or `assert` — something with a message —, yes, maybe using libvisual's `log_and_exit`?

kaixiong: @hartwork I've added a critical log message followed by `abort()`.

I spent some time thinking about the general approach of a portable `unreachable()` function that defaults to log-and-abort in debug builds... but I've ultimately concluded that for now, we should keep the change small for this PR 😄

hartwork (approved):

PR #384, opened 2025-01-23
Add LV::Palette methods for iterations and comparisons

kaixiong:
This PR implements:
- standard C++ container methods such as `begin() `and `end()` for iteration
- equality and inequality operators for comparing palettes

### DIFF
--- a/libvisual/libvisual/lv_palette.h
+++ b/libvisual/libvisual/lv_palette.h
@@ -45,6 +45,9 @@ namespace LV {
   //!
   struct LV_API Palette
   {
+      using const_iterator = std::vector<Color>::const_iterator;
+      using iterator = std::vector<Color>::iterator;
+
       std::vector<Color> colors;
 
       /**
@@ -78,6 +81,36 @@ namespace LV {
       constexpr friend bool operator!= (Palette const& lhs, Palette const& rhs)
       {
           return !(lhs == rhs);
+      }
+
+      constexpr const_iterator cbegin () const noexcept
+      {
+          return colors.cbegin ();
+      }
+
+      constexpr const_iterator cend () const noexcept
+      {
+          return colors.cend ();
+      }
+
+      constexpr const_iterator begin () const noexcept
+      {
+          return colors.begin ();
+      }
+
+      constexpr const_iterator end () const noexcept
+      {
+          return colors.end ();
+      }
+
+      constexpr iterator begin () noexcept
+      {
+          return colors.begin ();
+      }
+
+      constexpr iterator end () noexcept
+      {
+          return colors.end ();
       }
 
       bool empty () const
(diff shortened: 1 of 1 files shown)


libvisual/libvisual/lv_palette.h
```diff
@@ -35,6 +35,7 @@
 
 #ifdef __cplusplus
 
+#include <algorithm>
```
hartwork, line 38: I'm likely missing something, but I do not see anything in the first commit that would need `<algorithm>` and also not in the second. What is `<algorithm>` used for/by in the pull request?

libvisual/libvisual/lv_palette.h
```diff
@@ -45,6 +46,9 @@ namespace LV {
   //!
   struct LV_API Palette
   {
+      using const_iterator = std::vector<Color>::const_iterator;
+      using iterator = std::vector<Color>::iterator;
```
hartwork, line 50: While these two seem to help readability further down in some regard, the fact that the color aspect — a key aspect of the return type — is hidden away seems (worse than hiding the vector away) and like a disservice to me. Maybe `const_color_iterator`? Is there a concept other than D.R.Y. and brevity here that I am missing?

hartwork (changes requested):

libvisual/libvisual/lv_palette.h
kaixiong: You're right, it's not used. I think I missed this when splitting the commit out.

libvisual/libvisual/lv_palette.h
kaixiong: Type erasure is kind of the point for generics. These type definitions are part of `LV::Palette`'s interface for generic programming (think protocols in Python duck typing) and are standard in C++ containers.

For example, imagine a generic function that works on some sequence of `LV::Color` objects. Say, computing the average colour intensity of the collection. The algorithm doesn't know or want to know if it's a `std::vector<Color>`, a `std::array<Color>`, a C array (`LV::Color *`), `LV::Palette` or some fancy structure that does size optimization tricks. All it cares about is iterating, calculate each color's intensity, accumulating and finally averaging in the end.

So in that function, when it needs to declare and instantiate an iterator for the container, it can declare `Container::iterator`, `Container::const_iterator` as its type, where `Container` could be any of the above.

hartwork: @kaixiong by `Container` you mean `Palette` I assume? I think I understand where you are going with this. I will approve the pull request and only removal of `<algorithm>` remains to be changed…

hartwork (approved):

libvisual/libvisual/lv_palette.h
kaixiong: @hartwork, hmm upon some more thinking, perhaps I should also provide typedefs for `value_type` (= Color) and maybe `reference` (Color &) and `const_reference` (Color const&). This also serves as documentation what the element type is when iterating over an LV::Palette instance. I don't really want to implement the full [SequenceContainer](https://en.cppreference.com/w/cpp/named_req/SequenceContainer) interface just yet though.

PR #385, opened 2025-01-23
Change PNG image loader to load 8-bit indexed images without format conversion 

kaixiong:
Currently, the PNG loader applies a libpng input transformation to convert 8-bit indexed colour images into 24-bit RGB images. This may be an undesirable step and inconsistent with what the BMP loader does.

This patch drops this and loads 8-bit indexed colour PNGs directly into 8-bit colour LV::Video objects. Note that 8-bit grayscale images continue to be converted to 24-bit RGB.


hartwork (approved):
@kaixiong looks alright to the point that I can understand without dedicated research of my own :+1:

PR #386, opened 2025-01-23
Fix BMP palette loading

kaixiong:
This PR fixes the palette loading in the BMP image loader.

The bug is due to the loader reading from the wrong file offset.

BMP files begin with a fixed size file header and a DIB (device-independent bitmap) header that has been continually extended over the years. According to [Wikipedia](https://en.wikipedia.org/wiki/BMP_file_format), there are 8 versions of the DIB header with differing sizes. They vary from the original 12 bytes to as large as 124 bytes in recent revisions. Each version subsumes the fields of previous versions. The colour table i.e. palette is found right after [1].

The loader currently reads part of the DIB header to get what it needs and proceeds to read the palette entries without accounting for the actual DIB header size. This patch fixes this by seeking to the end of the DIB header, utilizing the size given by the header's first field.

[1] Except when the compression schemes `BI_BITFIELDS` or `BI_ALPHABITFIELDS` are used, which isn't supported by the loader.

### DIFF
--- a/libvisual/libvisual/private/lv_video_bmp.cpp
+++ b/libvisual/libvisual/private/lv_video_bmp.cpp
@@ -341,7 +341,7 @@ namespace LV {
           return nullptr;
       }
 
-      if (bi_compression > 3) {
+      if (bi_compression >= 3) {
           visual_log (VISUAL_LOG_ERROR, "Bitmap uses an invalid or unsupported compression scheme");
           fp.seekg (saved_stream_pos);
           return nullptr;
--- a/libvisual/libvisual/private/lv_video_bmp.cpp
+++ b/libvisual/libvisual/private/lv_video_bmp.cpp
@@ -286,6 +286,8 @@ namespace LV {
       fp.read (reinterpret_cast<char*> (&bf_bits), 4);
       bf_bits = VISUAL_ENDIAN_LEI32 (bf_bits);
 
+      auto dib_header_pos = fp.tellg ();
+
       /* Read the info structure size */
       fp.read (reinterpret_cast<char*> (&bi_size), 4);
       bi_size = VISUAL_ENDIAN_LEI32 (bi_size);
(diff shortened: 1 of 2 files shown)


libvisual/libvisual/private/lv_video_bmp.cpp
```diff
@@ -341,12 +343,14 @@ namespace LV {
           return nullptr;
       }
 
-      if (bi_compression > 3) {
+      if (bi_compression >= 3) {
```
hartwork, line 346: I'm not a big fan of approach "undocumented unnamed magic number". This seems to be 3 because `BI_BITFIELDS == 3` in Windows. If we cannot use that constant without including Windows headers, let's either (a) add a good comment or (b) make our own copy of that constant (or both)?

libvisual/libvisual/private/lv_video_bmp.cpp
```diff
@@ -341,12 +343,14 @@ namespace LV {
...
           visual_log (VISUAL_LOG_ERROR, "Bitmap uses an invalid or unsupported compression scheme");
           fp.seekg (saved_stream_pos);
           return nullptr;
       }
 
+      fp.seekg (dib_header_pos + std::streampos {bi_size}, std::ios::beg);
```
hartwork, line 352: Let's add comments that (or make otherwise more obvious)…
- that we're jumping to the start of the palette here (— we could also move the line_below_ comment `/* Load the palette */` to better show it's part of palette loading?)
- that this is the correct value (only) because we ruled out both `BI_BITFIELDS` and `BI_ALPHABITFIELDS` earlier (referring to the first paragraph at https://en.wikipedia.org/wiki/BMP_file_format#Color_table)

hartwork (changes requested):

libvisual/libvisual/private/lv_video_bmp.cpp
kaixiong: Good idea, let me make that change.

hartwork (approved):

PR #392, opened 2025-02-24
[0.4.x] Accept sdl12-compat as SDL 1.2.x dependency

marcin-serwin:
This makes it possible to build the examples and `lv-tool` on systems that do not provide SDL 1.2 but provide the compatibility layer.


hartwork (changes requested):
Hi @marcin-serwin, file `sdl12_compat.pc` comes with line `Provides: sdl = 1.2.68` so this should work out of the box for both sdl1 and sdl2-compat and it does in e.g. Gentoo. What distro is this with and do you not have the `Provides` line? Is packaging patching it out downstream?

hartwork:
@marcin-serwin thanks for the debugging and the detailed reply! :+1:

PR #412, opened 2026-01-21
po: Add Georgian translation

EkaterinePapava:


hartwork:
@EkaterinePapava thanks for the pull request! I have a few other things to do first, but I will have a closer look! :+1:

